PR #6, opened 2020-04-13
upgrade to 3.11

nathangoulding:
Upgrading our installer from Alpine 3.7 to Alpine 3.11, and testing the automatic rebuild of Alpine's initramfs, kernel, and modules if those bits change.

### DIFF
--- a/installer/alpine/build.sh
+++ b/installer/alpine/build.sh
@@ -2,17 +2,19 @@
 
 set -o errexit -o nounset -o pipefail -o xtrace
 
+arch=$(uname -m)
+
 # shellcheck disable=SC2039
-[[ $(uname -m) == aarch64 ]] && echo "aarch64 isn't _really_ tested/supported yet" && exit 1
+[[ "${arch}" == aarch64 ]] && echo "aarch64 isn't _really_ tested/supported yet" && exit 1
 
 # Install the eclypsiumdriver package previously built in the Dockerfile
-apk add --no-scripts --no-cache --update --upgrade /home/builder/packages/non-free/x86_64/eclypsium*.apk
+#apk add --no-scripts --no-cache --update --upgrade /home/builder/packages/non-free/x86_64/eclypsium*.apk
 
 # Eclypsium driver and supporting modules (IPMI)
-cat >/etc/mkinitfs/features.d/eclypsium.modules <<EOF
-extra/eclypsiumdriver.ko
-kernel/drivers/char/ipmi/*.ko
-EOF
+#cat >/etc/mkinitfs/features.d/eclypsium.modules <<EOF
+#extra/eclypsiumdriver.ko
+#kernel/drivers/char/ipmi/*.ko
+#EOF
 
 cat >/etc/mkinitfs/features.d/network.modules <<EOF
 kernel/drivers/net/ethernet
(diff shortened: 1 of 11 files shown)


nathangoulding:
Tests are green, and we're ready for initial code review. A few things before merging:

- We need to understand what we're doing with the Eclypsium module. I have an email thread going with the team. I'd prefer to fetch and add this in a more standard way, or remove it from osie and turn this into a workflow action. Thoughts here are welcome, but in all cases we'll need to address this.

- We will need to test this across our fleet to check backwards compatibility with all our systems. Any suggestions on how we can do this?

mmlb:
>     * We need to understand what we're doing with the Eclypsium module. I have an email thread going with the team. I'd prefer to fetch and add this in a more standard way, or remove it from osie and turn this into a workflow action. Thoughts here are welcome, but in all cases we'll need to address this.

Workflow makes a lot of sense to me.

>     * We will need to test this across our fleet to check backwards compatibility with all our systems. Any suggestions on how we can do this?

Well since you are in a deep-dive groove want to resurrect the provision based tests? It's a bit hard to follow but can be simplified a lot now that api has batch endpoint, it cleans up a lot of the logic. We could also test preinstalls this way too.

installer/runner.sh
```diff
@@ -81,9 +81,17 @@ phone_home_url=$(jq -r .phone_home_url "$metadata")
...
-	sleep 3
+for i in $(seq 5); do
+	echo "Waiting for docker to respond (${i})..." &&
+		docker info &&
+		break
+	sleep 5
 done
+if ! docker info; then
+	cat /var/log/docker.log
+	echo "Docker not responding, dropping to a shell..."
```
rgl, line 92: is this installer/runner.sh file manually executed by a human?

is it automatically executed by a machine? if so, this should not drop into a shell.

installer/alpine/build.sh
```diff
@@ -31,15 +33,15 @@ mv /etc/mkinitfs/features.d/virtio.modules.tmp /etc/mkinitfs/features.d/virtio.m
 
 # Make initramfs with features we think are spiffy
 # shellcheck disable=SC2016
-echo 'features="base ext2 ext3 ext4 keymap network packetrepo squashfs virtio eclypsium"' >/etc/mkinitfs/mkinitfs.conf
+echo 'features="base ext2 ext3 ext4 keymap network packetrepo squashfs virtio"' >/etc/mkinitfs/mkinitfs.conf
 kver=$(basename /lib/modules/*)
 mkinitfs -l "$kver"
-mkinitfs -o /assets/initramfs-vanilla "$kver"
+mkinitfs -o /assets/initramfs-"$arch" "$kver"
```
rgl, line 39: this is an odd way to quote a variable. isn't it nicer to the reader to use `"/assets/initramfs-$arch"` instead?

installer/alpine/init-x86_64
```diff
@@ -542,7 +644,7 @@ fi
 if [ -f "$ovl" ]; then
 	ebegin "Loading user settings from $ovl"
 	# create apk db and needed /dev/null and /tmp first
-	apk add --root $sysroot --initdb --quiet --no-progress
+	apk add --root $sysroot --initdb --quiet
```
rgl, line 647: why remove `--no-progress`? will apk to the right thing of not showing the progress when run in CI?

rgl (commented):
Bump this PR to Alpine 3.12 too?

mmlb:
> Bump this PR to Alpine 3.12 too?

maybe, one of the last bits of work here to verification on a bunch of our machines. Will see about updating first to and be able to knock both things out.

installer/runner.sh
mmlb: Runs automatically, drops to a shell when an issue happens and support and/or devs are able to poke around and debug.

installer/alpine/build.sh
mmlb: yes

installer/runner.sh
rgl: Isn't this executed in a CI environment? Without user access.

mrmrcoleman:
Hey @mmlb any update on this?

mmlb:
closing, since we'll be going with 3.12 and have further work to do.

PR #8, opened 2020-04-15
Remove c3.medium from kexec bypass

dustinmiller:
We opted for BIOS over UEFI for this plan, so we can kexec in now

- [x] Changelog updated


docker/scripts/osie.sh
```diff
@@ -425,7 +425,7 @@ EOF_tty
 	# Backup phone-home
 
 	# Add phone-home to OS
-	nonsystemdoses=(ubuntu_14_04 ubuntu_14_04_image scientific_6)
+	nonsystemdoses=(_14_04 ubuntu_14_04_image scientific_6)
```
ScottGarman: To my untrained eye, I'm not clear on why this change was made - can you explain? Doesn't need a code comment, I'm just looking to learn.

dustinmiller: This was unintentional, great catch

ScottGarman:
:+1:

mmlb (approved):

PR #11, opened 2020-04-15
Enable kexec bypass for centos_8 on t1.small.x86

ScottGarman:
Fix for ENG-4916.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) t
 ## Unreleased
 
 ### Added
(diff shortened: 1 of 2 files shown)


ScottGarman:
I'm not sure yet now I could verify this fix works, but @dlaube strongly suspects that this the root cause. Also, lemme know if I should leave out jira ticket references in commits, since that project isn't public and this is an open source repo.

dlaube (dismissed):
LGTM!

ScottGarman:
I've refactored this to work with Manny's recent changes to the reboot script, should be good to go.

mmlb (approved):

dustinmiller (approved):

PR #12, opened 2020-04-16
Tweak kexec/reboot handling in osie.sh

mmlb:
Adding some echo and set -x so we can see what is going on in the
cleanup.sh scripts for a bit better debugging. I also changed from a
bunch of if/elifs to a switch statement since its easier to read/grok.

- [ ] Changelog updated


ScottGarman:
I'm :+1: on the code changes, but not sure what's going on with CI - that may need to be addressed first.

mmlb:
I fixed the `if $reboot` so it won't bomb out when we not set due to `set -u`.

ScottGarman (approved):

PR #15, opened 2020-04-16
Remove c3.medium from chroot_install

dustinmiller:
- [ ] Changelog updated

### DIFF
--- a/docker/scripts/grub-installer.sh
+++ b/docker/scripts/grub-installer.sh
@@ -70,7 +70,7 @@ echo "OS: $DOS  ARCH: $arch VER: $DVER"
 
 chroot_install=false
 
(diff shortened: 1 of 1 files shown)


dustinmiller:
I don't think was doing much, but since it's not doing anything we should remove it?

ScottGarman:
My understanding is that this got lost during a prior migration of the repo. I'm :+1: on it, but not confident enough about the implications of this that I'll wait for Manny, David, or someone else to do a final review/merge.

mmlb (changes requested):
see commit comment

dustinmiller:
Looks like this was doing stuff and needs removed, updated commit description per @mmlb's request

dustinmiller:
Updated again, should be more descriptive

mmlb:
Thats the stuff!

mmlb (approved):

PR #19, opened 2020-04-17
Include Ub16+c3.med in the kexec bypass

dustinmiller:
These were stalling out at the end of the provision and failing.

- [x] Changelog updated


docker/scripts/osie.sh
```diff
@@ -539,6 +539,7 @@ echo -e "${BYELLOW}Install time: $((etimer - stimer))${NC}"
 case ${OS}:${class} in
 ubuntu_18_04:t1.small.x86) reboot=true ;;
 ubuntu_16_04:t3.small.x86) reboot=true ;;
+ubuntu_16_04:c3.medium.x86) reboot=true ;;
```
mmlb: sort!

mmlb (changes requested):

dustinmiller:
Will be doing squash here

PR #23, opened 2020-04-20
installer/workflow-helper-rc: add command_background=true

### LINKED ISSUE #22: workflow-helper script should not block TTY from spawning
It seems that right now, if `workflow-helper` script hits infinite loop trying to log in into the docker registry, there is no way to debug it and there is no logs produced to TTY either.

If you look at the serial console output, then the login prompt does not show up until `workflow-helper` script finishes execution.

invidian:
This commit adds command_background=true option to workflow-helper
openrc file, so it will run in the background instead of blocking
execution process.

This will unblock spawning the TTY process, which might be helpful while
debugging, as currently the workflow-helper script may run infinitely,
which causes TTY to never spawn.

Closes #22

Signed-off-by: Mateusz Gozdek <mateusz@kinvolk.io>

- [ ] Changelog updated

### DIFF
--- a/installer/workflow-helper-rc
+++ b/installer/workflow-helper-rc
@@ -2,3 +2,5 @@
 
 name="Packet Workflow"
 command="/sbin/workflow-helper"
+command_background=true
+pidfile=/run/workflow-helper.pid


invidian:
Should be working now.

gauravgahlot:
LGTM
@invidian Just wanted to confirm if you have tested the fix by intentionally making `workflow-helper`  fall in an infinite loop?

invidian:
>@invidian Just wanted to confirm if you have tested the fix by intentionally making workflow-helper fall in an infinite loop?

Yes, If I start some long-running workflow, which takes several minutes, I still get TTY login screen quickly (after 1 minute, but I think this is just my setup), so it's definitely not blocked infinitely.

Also, with my testing, I added some logging capabilities to the `workflow-helper` script, and I can log in into the machine and run `tail -f` to monitor the execution of the workflow.

gauravgahlot (approved):
That sounds great. Approved.

PR #25, opened 2020-04-20
installer/workflow-helper.sh: don't use -i in docker run

### LINKED ISSUE #21: workflow-helper.sh doesn't execute 2nd docker run
Hey there,

It looks like there is some sort of race condition in `workflow-helper.sh`. When `workflow-helper.sh` is executed as part of init.rd the following command from the mentioned script is not executed:

https://github.com/tinkerbell/osie/blob/master/installer/workflow-helper.sh#L69-L81
```
docker run --privileged -ti \
	-e "container_uuid=$id" \
	-e "WORKER_ID=$worker_id" \
	-e "DOCKER_REGISTRY=$docker_registry" \
	-e "TINKERBELL_GRPC_AUTHORITY=$grpc_authority" \
	-e "TINKERBELL_CERT_URL=$grpc_cert_url" \
	-e "REGISTRY_USERNAME=$registry_username" \
	-e "REGISTRY_PASSWORD=$registry_password" \
	-v /worker:/worker \
	-v /var/run/docker.sock:/var/run/docker.sock \
	--log-driver=fluentd -t \
	--net host \
	"$docker_registry/tink-worker:latest"
```
Re-running `workflow-helper.sh` manually succeeds and executes without problems.

invidian:
It seems running multiple containers with -i does not work when executed
from openrc service and causes 2nd container to fail to start.

It also doesn't make much sense to run with -i, as those are not really
an interactive sessions.

-t stays for both commands, if one wants to use 'docker attach' for some
reason.

Closes #21

Signed-off-by: Mateusz Gozdek <mateusz@kinvolk.io>

- [ ] Changelog updated

### DIFF
--- a/installer/workflow-helper.sh
+++ b/installer/workflow-helper.sh
@@ -57,7 +57,7 @@ done
 # stop mdev from messing with us once and for all
 rm -f /sbin/mdev
 
-docker run -dit --net host \
+docker run -dt --net host \
 	"$docker_registry/fluent-bit:1.3" \
 	/fluent-bit/bin/fluent-bit -i forward -o "es://$elastic_search_url/worker/worker"
 
@@ -66,7 +66,7 @@ sleep 3
 
 mkdir /worker
 
-docker run --privileged -ti \
+docker run --privileged -t \
 	-e "container_uuid=$id" \
 	-e "WORKER_ID=$worker_id" \
 	-e "DOCKER_REGISTRY=$docker_registry" \


parauliya:
Hi @invidian ,
This does make sense. But did you test this after removing the "-i" flag if yes, please explain in brief.

invidian:
>This does make sense. But did you test this after removing the "-i" flag if yes, please explain in brief.

Hey @parauliya, yes, I tested it. I'm currently testing https://github.com/tinkerbell/tink/pull/62 extensively. With that, I could modify the script on the fly in `/packet/nginx/workflow/` directory and test it many times. I've tried various combinations, including both adding and removing `-t` and `-d` parameters, and this I think makes the most sense, as I explained in the commit message.

parauliya:
> > This does make sense. But did you test this after removing the "-i" flag if yes, please explain in brief.
> 
> Hey @parauliya, yes, I tested it. I'm currently testing [tinkerbell/tink#62](https://github.com/tinkerbell/tink/pull/62) extensively. With that, I could modify the script on the fly in `/packet/nginx/workflow/` directory and test it many times. I've tried various combinations, including both adding and removing `-t` and `-d` parameters, and this I think makes the most sense, as I explained in the commit message.

That's sounds good. Changes are LGTM. I'll be approving this PR soon.

parauliya (approved):
Changes are LGTM.

PR #34, opened 2020-04-28
Adjust OSIE logging

dlaube:
- [ ] Changelog updated


ScottGarman:
Changes look good, but maybe also mention in the commit log why removing xtrace is desired, for better context?

ScottGarman (approved):

PR #35, opened 2020-05-11
Prefer our local github-mirror for packet-images repo access

ScottGarman:
We've set up a mirror of our packet-images repo in order to prevent
provisioning failures when GitHub is unavailable/throttled. Downloads
happen via our proxy cache at images.packet.net.

### DIFF
--- a/docker/scripts/functions.sh
+++ b/docker/scripts/functions.sh
@@ -664,6 +664,34 @@ function ensure_reachable() {
 	echo -e "${YELLOW}###### OK${NC}"
 }
 
+# Check that our git-lfs mirror server is working
+function github_mirror_check() {
+	echo -e "${YELLOW}###### Checking the health of github-mirror.packet.net...${NC}"
+
+	# Clone the repo, and checkout an LFS branch.
+	local timeout=30 # seconds
+	local lfs_testing_uri="https://github-mirror.packet.net/packethost/lfs-testing.git"
+	local lfs_testing_branch="remotes/origin/images-tiny"
+	if ! timeout --preserve-status $timeout git clone -q $lfs_testing_uri; then
+		echo -e "${YELLOW}###### Timeout when cloning the lfs-testing repo${NC}"
+		return 1
+	fi
+	cd lfs-testing
+	if ! timeout --preserve-status $timeout git checkout -q $lfs_testing_branch; then
+		cd .. && rm -rf lfs-testing
+		echo -e "${YELLOW}###### Timeout when checking out git-lfs data from lfs-testing${NC}"
+		return 1
+	fi
+	checksum=$(sha256sum tiny/tiny.img | awk '{ print $1 }')
+	cd .. && rm -rf lfs-testing
+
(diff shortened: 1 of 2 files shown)


ScottGarman:
This is ready for another round of feedback, especially on how I'm checking the health of github-mirror and my process of falling back to github.com if that fails.

docker/scripts/functions.sh
```diff
@@ -664,6 +664,33 @@ function ensure_reachable() {
...
+function github_mirror_check() {
+	echo "Checking the health of github-mirror.packet.net"
+	is_reachable "github-mirror.packet.net" || return 1
+
+	# Clone the repo, and checkout an LFS branch.
+	local timeout=30 # seconds
+	local lfs_testing_uri="https://github-mirror.packet.net/packethost/lfs-testing.git"
+	local lfs_testing_branch="remotes/origin/images-tiny"
+	timeout --preserve-status $timeout \
+		git clone $lfs_testing_uri >/dev/null 2>&1 ||
```
mmlb: why not git clone -q and drop the redirects?

docker/scripts/functions.sh
```diff
@@ -664,6 +664,33 @@ function ensure_reachable() {
...
+		(echo "Timeout when cloning the lfs-testing repo" && return 1)
+	cd lfs-testing
+	timeout --preserve-status $timeout \
+		git checkout $lfs_testing_branch >/dev/null 2>&1 ||
```
mmlb: ditto re: -q

docker/scripts/osie.sh
```diff
@@ -130,11 +130,18 @@ if ! [[ -f /statedir/disks-partioned-image-extracted ]]; then
 	mkdir $assetdir
 	echo -e "${GREEN}#### Fetching image (and more) via git ${NC}"
 
-	# config hosts entry so git-lfs assets are pulled through our image cache
-	githost="images.packet.net"
-	images_ip=$(getent hosts $githost | awk '{print $1}')
+	# config hosts entry so git-lfs assets from github and our github-mirror are
+	# pulled through our image cache
+	githost="github-mirror.packet.net"
+	if ! github_mirror_check; then
```
mmlb: will this fail when run in pre-install mode and dev has the non-routable 172/16 ip? See call to  `ensure_reachable` on line 164.

mmlb (changes requested):

docker/scripts/osie.sh
ScottGarman: Good catch. I've updated the github_mirror_check() function to include some of that logic and handle that situation now. I've also noticed I wasn't using colors in my warning messages and have updated those accordingly.

mmlb: I don't feel too great about extracting some of the ensure_reachable logic into github_mirror_check, why not do:
```sh
if ! (ensure_reachable github-mirror.packet.net && github_mirror_check) ; then
...
```

mmlb (dismissed):

dlaube (approved):
Nice. This looks pretty good to me!

PR #37, opened 2020-05-19
Read eclypsium token from kernel cmdline | ENG-5336

splaspood:
Provide an eclypsium scanning token during runtime to avoid having to hard code
    it into osie.

    * Parse eclypsium_token= from /proc/cmdline
    * Pass eclypsium_token to osie 'mode' script as env ECLYPSIUM_TOKEN
    * Utilize passed token in EclypsiumApp scanner invocation

### DIFF
--- a/docker/scripts/deprovision.sh
+++ b/docker/scripts/deprovision.sh
@@ -207,19 +207,15 @@ baremetal_2a2 | baremetal_2a4 | baremetal_2a5 | baremetal_hua)
 esac
 
 # Run eclypsium
-if [[ $arch == x86_64 ]]; then
-	case "$facility" in
-	ssg-* | att-*)
-		echo "skipping eclypsium in unsupported facility"
-		;;
-	*)
+if [[ -n "${ECLYPSIUM_TOKEN:-}" ]]; then
+	if [[ $arch == x86_64 ]]; then
 		case "$class" in
 		disabled.plan.here)
 			echo "skipping eclypsium on unsuppported plan"
 			;;
 		*)
 			https_proxy="http://eclypsium-proxy-${facility}.packet.net:8888/" /usr/bin/EclypsiumApp \
-				-s1 prod-0918.eclypsium.net placeholder \
+				-s1 prod-0918.eclypsium.net "${ECLYPSIUM_TOKEN}" \
 				-disable-progress-bar \
 				-medium \
 				-log stderr \
@@ -227,8 +223,7 @@ if [[ $arch == x86_64 ]]; then
 				-custom-id "${id}" || echo 'EclypsiumApp Failed!'
 			;;
 		esac
-		;;
-	esac
+	fi
 fi
 
 phone_home "${tinkerbell}" '{"type":"deprovisioning.306.02","body":"Deprovision finished, rebooting server","private":true}'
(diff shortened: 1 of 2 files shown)


installer/osie-installer.sh
```diff
@@ -173,6 +174,7 @@ docker run --privileged -ti \
 	-h "${hardware_id}" \
 	-e "container_uuid=$id" \
 	-e "RLOGHOST=$tinkerbell" \
+	-e "ECLYPSIUM_TOKEN=$eclypsium_token" \
```
mmlb: this needs to be `${eclypsium_token:-}` otherwise will cause an error when using an unset variable right?

mmlb (changes requested):

installer/osie-installer.sh
splaspood: The sed parsing of /proc/cmdline will fail to define the env var if eclypsium_token is not present, regardless if it has an empty value or not.    Subsequently when docker run is called, the '-e ECLYPSIUM_TOKEN=' parameter will not cause an error, and will pass the 'empty' variable into the container.    That being said, I agree with being more explicit, so I have made the suggested change.

docker/scripts/deprovision.sh
```diff
@@ -207,28 +207,30 @@ baremetal_2a2 | baremetal_2a4 | baremetal_2a5 | baremetal_hua)
...
-if [[ $arch == x86_64 ]]; then
-	case "$facility" in
-	ssg-* | att-*)
-		echo "skipping eclypsium in unsupported facility"
-		;;
-	*)
-		case "$class" in
-		disabled.plan.here)
-			echo "skipping eclypsium on unsuppported plan"
+if [ -z "$ECLYPSIUM_TOKEN" ] && [ "${ECLYPSIUM_TOKEN}" != "" ]; then
```
mmlb: `if [[ -n "${ECLYPSIUM_TOKEN:-}" ]]`. You need the `${...:-}` because even with `-z` you are still using a possibly undefined variable and bash stops you before it runs the `-z` test if it is in fact undefined.

Then also the `if` conditions don't make sense as is anyway. `if empty && != ""` so will never match.

docker/scripts/deprovision.sh
```diff
@@ -207,28 +207,30 @@ baremetal_2a2 | baremetal_2a4 | baremetal_2a5 | baremetal_hua)
...
+	if [[ $arch == x86_64 ]]; then
```
mmlb, line 211: This test condition can be folded into the `case "$class" in` statement below. Instead of `*` you can do `*.x86.*)`. Would be nice to remove logic nesting/indentation.

mmlb (changes requested):

docker/scripts/deprovision.sh
splaspood: Ya know, my 'entire life' I've thought -z meant something different than it turns out it actually does...     Changes made as suggested.

docker/scripts/deprovision.sh
splaspood: There are plans without x86 in them still, so I wouldn't combine that with the plan check for that reason.

docker/scripts/deprovision.sh
mmlb: heh

docker/scripts/deprovision.sh
mmlb: sigh :(. Do we have a ticket to convert over?

mmlb: wait plans maybe but class?

mmlb (approved):

docker/scripts/deprovision.sh
splaspood: Class?  plan versions?

PR #44, opened 2020-06-02
OS image git fetch improvements

ScottGarman:
Since we only need the state of the images repo at a particular commit, we can avoid downloading the entire repo metadata by doing a shallow fetch with a depth of 1. This should substantially reduce our OS deployment times.

This also increases the git lfs config setting lfs.transfer.maxtretries to 10 in an attempt to prevent intermittent "smudge filter" failures we occasionally see with lfs checkouts from GitHub.


ScottGarman:
I've tested the new git workflow in an osie container running locally, and confirmed that it works with both GitHub and our github-mirror servers.

mmlb (approved):

PR #45, opened 2020-06-02
Prefer our local github-mirror for packet-images repo access

ScottGarman:
We've set up a mirror of our packet-images repo in order to prevent
provisioning failures when GitHub is unavailable/throttled. Prefer
github-mirror.packet.net, and fallback to github.com if health checks
fail for github-mirror.

### DIFF
--- a/docker/scripts/osie.sh
+++ b/docker/scripts/osie.sh
@@ -130,16 +130,30 @@ if ! [[ -f /statedir/disks-partioned-image-extracted ]]; then
 	mkdir $assetdir
 	echo -e "${GREEN}#### Fetching image (and more) via git ${NC}"
 
-	# config hosts entry so git-lfs assets are pulled through our image cache
-	githost="images.packet.net"
-	images_ip=$(getent hosts $githost | awk '{print $1}')
+	# config hosts entry so git-lfs assets from github and our github-mirror are
+	# pulled through our image cache
+	images_ip=$(getent hosts images.packet.net | awk '{print $1}')
 	cp -a /etc/hosts /etc/hosts.new
-	echo "$images_ip        github-cloud.s3.amazonaws.com" >>/etc/hosts.new && cp -f /etc/hosts.new /etc/hosts
+	{
+		echo "$images_ip        github-cloud.s3.amazonaws.com"
+		echo "$images_ip        github.com"
+		echo "$images_ip        github-mirror.packet.net"
+	} >>/etc/hosts.new
+	# Note: using mv here fails (415 Unsupported Media Type) because docker sets
+	# this up as a bind mount and we can't replace it.
+	cp -f /etc/hosts.new /etc/hosts
 	echo -n "LFS pulls via github-cloud will now resolve to image cache:"
 	getent hosts github-cloud.s3.amazonaws.com | awk '{print $1}'
 
 	if [[ ${OS} =~ : && $custom_image == false ]]; then
 		image_tag=$(echo "$OS" | awk -F':' '{print $2}')
+
+		githost="github-mirror.packet.net"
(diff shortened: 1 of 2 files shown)


dustinmiller (dismissed):
Looks good :crossed_fingers:

ScottGarman:
Note: This can't be merged or deployed until we have a plan for ipn on ewr1.

docker/scripts/osie.sh
```diff
@@ -130,17 +130,29 @@ if ! [[ -f /statedir/disks-partioned-image-extracted ]]; then
...
-	images_ip=$(getent hosts $githost | awk '{print $1}')
+	# config hosts entry so git-lfs assets from github and our github-mirror are
+	# pulled through our image cache
+	images_ip=$(getent hosts images.packet.net | awk '{print $1}')
 	cp -a /etc/hosts /etc/hosts.new
-	echo "$images_ip        github-cloud.s3.amazonaws.com" >>/etc/hosts.new && cp -f /etc/hosts.new /etc/hosts
+	{
+		echo "$images_ip        github-cloud.s3.amazonaws.com"
+		echo "$images_ip        github.com"
+		echo "$images_ip        github-mirror.packet.net"
```
mmlb, line 140: This can be just one line right?

docker/scripts/osie.sh
```diff
@@ -130,17 +130,29 @@ if ! [[ -f /statedir/disks-partioned-image-extracted ]]; then
...
+	} >>/etc/hosts.new
+	cp -f /etc/hosts.new /etc/hosts
```
mmlb, line 144: `mv` actually is better here.

mmlb (dismissed):

docker/scripts/osie.sh
ScottGarman: I seem to recall having adopted this formatting due to complains from shfmt. To otherwise do this with a single line would probably be an excessively long line.

docker/scripts/osie.sh
ScottGarman: I'll fix this.

mmlb (dismissed):

docker/scripts/osie.sh
mmlb: I guess I lied, `mv` is definitely not better here.

docker/scripts/osie.sh
```diff
@@ -130,17 +130,30 @@ if ! [[ -f /statedir/disks-partioned-image-extracted ]]; then
...
+	# pulled through our image cache
+	images_ip=$(getent hosts images.packet.net | awk '{print $1}')
 	cp -a /etc/hosts /etc/hosts.new
-	echo "$images_ip        github-cloud.s3.amazonaws.com" >>/etc/hosts.new && cp -f /etc/hosts.new /etc/hosts
+	{
+		echo "$images_ip        github-cloud.s3.amazonaws.com"
+		echo "$images_ip        github.com"
+		echo "$images_ip        github-mirror.packet.net"
+	} >>/etc/hosts.new
+	# Note: using mv here fails (415 Unsupported Media Type)
```
mmlb: I know why this fails though, its because its bind mounted into the container so we can't move it. So maybe lets make the wording something like:

```
# Note: using mv here fails (415 Unsupported Media Type) because docker sets this up as a bind mount and we can't replace it.
```

mmlb (changes requested):

docker/scripts/osie.sh
ScottGarman: Ah, I was just reading about this error and wondering why it was showing up in client/server error explanations. Will do.

mmlb (approved):

PR #48, opened 2020-06-09
Use openRC depend rather than executing "service docker start" from "workflow-helper" script / Log workflow-helper output into the file

### LINKED ISSUE #24: Use openRC depend rather than executing "service docker start" from "workflow-helper" script
https://github.com/OpenRC/openrc/blob/master/service-script-guide.md#the-depend-function

### LINKED ISSUE #26: Log workflow-helper output into the file
So it can be checked after the execution for investigating.

RafaelAnsay:
Fixed #24 
Also Fixed #26 
- [ ] Changelog updated

### DIFF
--- a/installer/workflow-helper.sh
+++ b/installer/workflow-helper.sh
@@ -42,13 +42,6 @@ wget "$packet_base_url/ca.pem"
 mkdir -p /etc/docker/ /etc/docker/certs.d/ "/etc/docker/certs.d/$docker_registry"
 cp ca.pem "/etc/docker/certs.d/$docker_registry/ca.crt"
 
-service docker start
-
-until docker info; do
-	echo 'Waiting for docker to respond...'
-	sleep 3
-done
-
 until docker login "$docker_registry" -u "$registry_username" -p "$registry_password"; do
 	echo 'Waiting for docker to respond...'
 	sleep 3
--- a/installer/workflow-helper-rc
+++ b/installer/workflow-helper-rc
@@ -4,3 +4,8 @@ name="Packet Workflow"
 command="/sbin/workflow-helper"
 command_background=true
 pidfile=/run/workflow-helper.pid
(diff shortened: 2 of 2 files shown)


RafaelAnsay:
depend() {need docker} in workflow-helper-rc dynamically starts docker

installer/workflow-helper.sh
```diff
@@ -42,6 +42,8 @@ wget "$packet_base_url/ca.pem"
 mkdir -p /etc/docker/ /etc/docker/certs.d/ "/etc/docker/certs.d/$docker_registry"
 cp ca.pem "/etc/docker/certs.d/$docker_registry/ca.crt"
 
+# wait for docker to fully start
+sleep 3
```
invidian: There is `until` loop right after it, which will retry if command fails. Is this sleep really needed?

invidian (changes requested):
Nice PR @RafaelAnsay. I'd squash first 2 commits, to avoid bloating git history. Other than that it looks good to me!

installer/workflow-helper.sh
RafaelAnsay: yes. in my dev environment docker login fails at least once because docker has not fully started.

invidian: But if there is a `until` loop, it should retry, so I think it's fine if it fails?

RafaelAnsay: sure i will remove the 1st sleep i added

parauliya:
Hey @RafaelAnsay ,
Have you tested this with the end to end workflow? If not please test this and post the result here.

RafaelAnsay:
sure i will post the results here

parauliya:
> sure i will post the results here

Hi @RafaelAnsay , what is the status of this PR? I think it has been in an open state for quite some time. Please provide the test results so that we can merge this PR asap.

PR #50, opened 2020-06-14
Upgrade git to 2.27 for improved shallow clone support

ScottGarman:
The version of git that comes with Xenial (2.7.4) doesn't support
shallow clones from our github-mirror server, so use the latest
officially supported Xenial release from the Ubuntu git maintainers PPA.


ScottGarman:
Side note: the osie package with this version of git is about 300 MB smaller than the previous one I've been using. A future optimization (to be done as separate work) could be to disable the building of several of the git packages (e.g, git-mediawiki, git-cvs, git-svn, git-gitk) which could improve our CI build times.

mmlb (approved):

Macgyverops (approved):

PR #51, opened 2020-06-16
Drop elasticsearch

grahamc:
re https://github.com/tinkerbell/tink/pull/170

The first commit is a bugfix, the second implements the real change. Should we add a step of forwarding the system's logs to boots' syslog server?


installer/workflow-helper.sh
```diff
@@ -57,13 +56,6 @@ done
 # stop mdev from messing with us once and for all
 rm -f /sbin/mdev
 
-docker run -dt --net host \
```
mmlb, line 60: we need to get rid of the `--log-driver` in the `docker run` call right?

mmlb (dismissed):

mmlb:
> Should we add a step of forwarding the system's logs to boots' syslog server?
That should be done already by the entrypoint.sh I believe.

mmlb (approved):

PR #54, opened 2020-06-22
Show Makefile some love

mmlb:
This Makefile has grown and grown, longer and more complicated. Time to reign it in, how else but with a layer of abstraction?!

This is mostly addressing the low hanging fruits in the Makefile. Things like moving the alpine init scripts into apps to disconnect from the boot files (for example, what does installer even mean wrt rescue?) I removed old stuff for fetching/verifying files that are no longer needed now that we use git-lfs for them. 

The big readability win was makefile macros -> jinja2 templates. This gives us a rendered output file that we can actually read, which we can't do with make's internal representation or even with `make -d`. We also get a much easier way to add support for new "platforms" (though I hope we get no more arm snowflakes) by just adding a new `platform:$arch` entry and regenerate.

There are more things needed to bring this Makefile out of the early-drone packet days and even getting rid of all pattern rules (once again for readability), but that will come in a different PR in the future.


mmlb:
This is a fairly big PR and I'm sorry about that, there are probably a couple of commits I could drop and do as a separate PR before this one, but I'm not sure its worth it.

Reading each commit (in git order, not github order!) might help a little bit.

ScottGarman (dismissed):
Totally dig these improvements. :100:

mrmrcoleman:
@grahamc to merge

grahamc:
_grits teeth, adds merge label_

grahamc (dismissed):

grahamc (approved):

PR #55, opened 2020-06-23
Always fetch metadata from using https

mmlb:
Metadata is looked up with https later in the script anyway so if https
fails it will now just fail sooner. This is needed for new deployments
where we are having issues with metadata on http.


mikemrm (dismissed):

mrmrcoleman:
Hey @mikemrm any update on this one?

mmlb:
This should be closed, we can't use https:// in `ensure_time` since its whole reason for existence is to hack around time being wrong and not having NTP, and if thats the case then using https will likely lead to cert not-before validity issues.

PR #57, opened 2020-06-24
look up hegel hosta and via SRV records

mikemrm:
Hegel doesn't really need to be on the metadata ip and it being on this
weird port is a pain in k8s. So lets use dns srv records to inform us
where to go, falling back to the old setup in case of any errors.

- [ ] Changelog updated


ScottGarman:
This lgtm but I'll let Manny do the final review. :+1:

installer/osie-installer.sh
```diff
@@ -14,10 +14,15 @@ fail() {
...
+#
+# This needs to be http because the whole reason for it is RTC might be wrong,
+# so if we were to try https in that scenario we'd get a certificate validity
+# error.
 ensure_time() {
 	local d hwdate mddate month
 	local months='jan feb mar apr may jun jul aug sep oct nov dec'
-	d=$(curl -sI http://metadata.packet.net/metadata | sed -n '/^Date:/ s|Date: ||p')
+	# *must* be http, not https
+	d=$(curl -sI h"ttp://tinkerbell.$facility.packet.net" | sed -n '/^Date:/ s|Date: ||p')
```
dlaube: typo on http near quote?

dlaube (changes requested):
Looks good I think! Found one issue that I commented on, otherwise I'd just make doubly sure that the python additions is well tested.

installer/osie-installer.sh
mikemrm: nice catch!

mikemrm:
> Looks good I think! Found one issue that I commented on, otherwise I'd just make doubly sure that the python additions is well tested.

Awesome, yep, runner changes have been tested!

mmlb:
we going to keep the `set -x`?

dlaube (approved):

PR #59, opened 2020-06-25
Add build-mstflint.sh

dustinmiller:
Older versions of mstflint weren't able to handle the latest Firmware 
version reporting

- [ ] Changelog updated


dlaube (approved):
Looks good!

PR #64, opened 2020-07-07
Gather firmware version and SMART attributes for NVMe

dustinmiller:


docker/scripts/hardware/utils.py
```diff
@@ -100,6 +100,8 @@ def get_smart_devices():
     for file in os.listdir("/dev"):
         if re.compile("^sd[a-z]$").match(file):
             disks.append(os.path.join("/dev", file))
+        if re.compile("^nvme[0-9]$").match(file):
+            disks.append(os.path.join("/dev", file))
```
zevweiss: While I don't see anything "wrong" with it per se, note that with the change below to remove the call to this function from `get_smart_diskprop()` I don't think this part would have any effect?

dustinmiller: I suspect this is the case as well, a remnant of some testing, removed

docker/scripts/hardware/utils.py
```diff
@@ -110,9 +110,15 @@ def get_smart_devices():
 
         dev, _, disk_type = line.split(" ")[0:3]
 
-        if "megaraid" not in disk_type:
+        if "megaraid" in disk_type:
```
zevweiss: Hmm, I'm a little confused here...we're going from megaraid-only to anything-but-megaraid on this path?

docker/scripts/hardware/utils.py
```diff
@@ -110,9 +110,15 @@ def get_smart_devices():
...
             continue
 
+        if "nvme" in disk_type:
+            smart_map[disks[counter] + "n1"] = disk_type
+        else:
+            smart_map[disks[counter]] = disk_type
```
zevweiss: I'm still confused by this counter-based business...it seems like we're relying on the relative ordering of what `os.listdir()` returns vs. what `smartctl --scan-open` prints and effectively sort of manually `zip()`-ing them together?  Except I think at least `os.listdir()`'s return wouldn't have any particular ordering guaranteed, so I'm not sure if it's just been working by luck or I'm misunderstanding what's going on...perhaps @splaspood can comment?

docker/scripts/hardware/component/disk.py
```diff
@@ -24,6 +24,9 @@ def __init__(self, lsblk):
         if not self.__is_nvme():
             self.data["smart"] = utils.get_smart_attributes(self.lsblk["name"])
 
+        if self.__is_nvme():
```
zevweiss, line 24: Given the two conditions involved, perhaps this and `if` above it could be converted to an `if`/`else`?

docker/scripts/hardware/utils.py
zevweiss: And further, with this version we don't have the regex tweak, so the `disks` list is still only going to have `sd*` entries, right?  So I'm not sure how the `+ "n1"` part could have done anything useful, since I don't see how `disks[counter]` could produce e.g. `"nvme0"`...

[Edit: no longer applicable after update adding the regex tweak.]

zevweiss: I guess perhaps my high-level question (for @splaspood I guess?) is why we're using `os.listdir()` at all instead of just using the device names from the smartctl output...

splaspood: If only I could recall...  I have to assume that at some point, or under certain conditions, smartctl wasn't presenting all devices...   but it's been so long since I wrote this that I no longer recall my reasoning.

zevweiss:
Nice...looks good to me!

ScottGarman (approved):

PR #65, opened 2020-07-07
Add support for Packet x.large.arm

scott4000:


docker/scripts/deprovision.sh
```diff
@@ -109,7 +109,9 @@ if [[ $preserve_data == false ]]; then
 	nvme_drives=($(find /dev -regex ".*/nvme[0-9]+" | sort -h))
 	echo "Found ${#nvme_drives[@]} nvme drives"
 	nvme list
-	if ((${#nvme_drives[@]} > 0)); then
+	if [[ $class == x.large.arm ]]; then
```
mikemrm, line 112: Can you add more details to the commit as to why were disabling this for this plan?

mikemrm (changes requested):

docker/scripts/deprovision.sh
mikemrm: Talking with Scott, this seems to be related to the ssds used in this plan lie about it's management capabilities

mikemrm (dismissed):

mikemrm (approved):

docker/scripts/deprovision.sh
scott4000: Additional details added to explain that the x.large.arm is in the evaluation phase and needs additional work to support NVMe namespace management.

PR #72, opened 2020-07-17
Alpine 3.12

### LINKED ISSUE #1: Outdated Alpine Linux version referenced
An outdated Alpine Linux version (3.7) is referenced in the README, and may also be in use within `osie`.

@nathangoulding it would be worth updating to harden and get the latest security patches.

I assume the built-in CI would validate that this change worked?

cc @vielmetti @gauravgahlot

mmlb:
## Description

Update our x86_64 alpine to v3.12, upgrade kernel to latest lts (5.4.52) and add Pensando NIC drivers. Also lift the rules from installer/alpine/Makefile into rules.mk.j2 so that the top-level `make` can handle building the alpine files itself. 

## Why is this needed

Fixes: #1 

## How Has This Been Tested?

Tested via the qemu based tests, also in production on various hardware.

Newer kernel means better hardware support. A supported Alpine base means bug fixes, especially ca-certificates updates :D.


installer/alpine/Dockerfile
```diff
@@ -17,56 +17,54 @@ RUN apk add --update --upgrade \
...
         linux-headers \
-        ;
-
-# remove edge when linux-vanilla is > 4.14.8 and has the nfp driver enabled,
-# likely alpine v3.8
-RUN apk add --no-scripts --no-cache --update --upgrade --repository http://dl-cdn.alpinelinux.org/alpine/edge/main \
+        && \
+    # remove edge when linux-vanilla is > 4.14.8 and has the nfp driver enabled,
+    # likely alpine v3.8
+    apk add --no-scripts --no-cache --update --upgrade --repository http://dl-cdn.alpinelinux.org/alpine/edge/main \
```
mikemrm: Can we see if this is no longer needed? I believe this was to get around some old packages.

rules.mk.j2
```diff
@@ -220,17 +221,19 @@ else
 	ci/vm.sh network_test | tee build/$@.log >/dev/$T
 endif
 
+build/osie-aarch64.tar.gz: /proc/sys/fs/binfmt_misc/qemu-aarch64
 build/osie-aarch64.tar.gz: SED=/FROM/ s|.*|FROM multiarch/ubuntu-debootstrap:arm64-xenial|
-build/osie-x86_64.tar.gz: SED=
+build/osie-x86_64.tar.gz:  SED=
```
mikemrm: extra space?

rules.mk.j2
```diff
@@ -220,17 +221,19 @@ else
...
 	$(E)"DOCKER   $@"
 	$(Q)sed '${SED}' $< > $<.$*
 	$(Q)docker build -t osie:$* -f $<.$* $(<D) 2>&1 | tee $@.log >/dev/$T
 	$(Q)docker save osie:$* > $@.tmp
 	$(Q)mv $@.tmp $@
 
+build/osie-runner-aarch64.tar.gz: /proc/sys/fs/binfmt_misc/qemu-aarch64
 build/osie-runner-aarch64.tar.gz: SED=/FROM/ s|.*|FROM multiarch/alpine:arm64-v3.7|
-build/osie-runner-x86_64.tar.gz: SED=
+build/osie-runner-x86_64.tar.gz:  SED=
```
mikemrm: ditto

rules.mk.j2
```diff
@@ -220,17 +221,19 @@ else
...
+build/osie-x86_64.tar.gz:  SED=
 build/osie-%.tar.gz: docker/Dockerfile ${osiesrcs}
 	$(E)"DOCKER   $@"
 	$(Q)sed '${SED}' $< > $<.$*
 	$(Q)docker build -t osie:$* -f $<.$* $(<D) 2>&1 | tee $@.log >/dev/$T
 	$(Q)docker save osie:$* > $@.tmp
 	$(Q)mv $@.tmp $@
 
+build/osie-runner-aarch64.tar.gz: /proc/sys/fs/binfmt_misc/qemu-aarch64
 build/osie-runner-aarch64.tar.gz: SED=/FROM/ s|.*|FROM multiarch/alpine:arm64-v3.7|
```
mikemrm: Do you want to wrap these in an if statement like line 180?

rules.mk.j2
mmlb: was on purpose for alignment ;p. Doesn't hurt.

rules.mk.j2
mmlb: nope because this is hard coded to aarch64, and so we don't need the if.

rules.mk.j2
mmlb: same

installer/alpine/Dockerfile
mmlb: the comment doesn't make any more sense but we still want latest and greatest linux-firmare. I tried building off of the pinned aports config but it wasn't working for me yesterday. Checked again today and it looks like the dev.alpinelinux.org host is back up again. Will use that then.

mmlb: Now installing linux-firmware from the checked out aports

mikemrm (dismissed):
lgtm!

mikemrm (approved):

ci/Dockerfile
```diff
@@ -1,39 +1,46 @@
...
+        python3-dev \
+        qemu-system-aarch64 \
+        qemu-system-x86_64 \
+        rsync \
+        socat \
+        util-linux \
+        && \
+    mv /etc/apk/repositories /etc/apk/repositories.old && \
+    apk add --no-cache --update --upgrade \
+        --repository=http://dl-cdn.alpinelinux.org/alpine/v3.11/main \
```
rgl, line 38: I think this needs to be changed to `v3.12` (maybe do a global search and replace just to check whether this does not appear elsewhere).

mmlb: no this is to get lshw from 3.11 explicitly. v3.12's `lshw` doesn't output correct json. I have not been able to track down the commit that broke `lshw`.

rgl: Oh I see :-(

mmlb: Its only used in tests so not a big deal imo. Looks like upstream has fixed the json output though so I'll look into PR'ing patches to aports.

mmlb: https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/11607 in case you're interested @rgl

PR #74, opened 2020-07-21
osie-runner: do not print exception tracebacks during hegel reconnects

ScottGarman:
When OSIE reconnects to hegel, an exception can be thrown and cause a
traceback to be displayed. This isn't really an error case and can cause
confusion for users trying to troubleshoot errors, so don't run the
exception logger, and rely on the info logger instead.

This is a fix for ENG-7075.

### DIFF
--- a/osie-runner/run.py
+++ b/osie-runner/run.py
@@ -120,7 +120,6 @@ while True:
     try:
         resp = watch.next()
     except grpc.RpcError as e:
-        log.exception("grpc error")
         log.info("hegel went away, attempting to reconnect")
         while True:
             try:


osie-runner/run.py
```diff
@@ -119,7 +119,7 @@ def connect_hegel(facility):
     log.info("about to monitor")
     try:
         resp = watch.next()
-    except grpc.RpcError as e:
+    except (grpc.RpcError, grpc._channel._Rendezvous) as e:
         log.exception("grpc error")
```
mikemrm: Hmm, I don't think this will change the level of verbosity. If an exception was being raise that wasn't being caught, it would fail out of this while loop and exit. Where as currently it prints the exception and retries.

I'm thinking perhaps this `log.exception` line is probably printing the exception along with the provided message. So if we're trying to get rid of the exception output, we might just need to remove that line.

mikemrm (changes requested):

ScottGarman:
I took a look at the Logger.exception documentation and agree, it's the source of the traceback. Since we still provide useful details to the user via the info logger line, I agree that simply removing the log.exception line should do the trick, and have updated this PR accordingly.

mikemrm (approved):

PR #75, opened 2020-07-21
Gather a better Mellanox model, from partNumber

dustinmiller:
## Description

For our Mellanox cards, lets gather partNumber for the model

## Why is this needed

Since we aren't doing this currently and we're reporting back name and model as the same thing. This is needed to pin firmware based on model which currently doesn't reflect reality well enough to do so.

## How Has This Been Tested?
Tested with the inventory.py, successfully posted

No impact expected

### DIFF
--- a/docker/scripts/hardware/utils.py
+++ b/docker/scripts/hardware/utils.py
@@ -253,6 +253,11 @@ def lspci(pci_id):
     return _lspci
 
 
+def get_mellanox_part_number(pci_id):
+    tree = etree.fromstring(mlxup_query(pci_id))
+    return tree[0].attrib["partNumber"]
+
+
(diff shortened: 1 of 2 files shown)


docker/scripts/hardware/utils.py
```diff
@@ -253,6 +253,11 @@ def lspci(pci_id):
     return _lspci
 
 
+def get_mellanox_part_number(pci_id):
+    tree = etree.fromstring(mlxup_query(pci_id))
+    return tree[0].attrib["partNumber"]
```
mikemrm, line 258: Could there be a point where the tree returns more than one item?

mikemrm (approved):

PR #78, opened 2020-07-31
Added container name for worker

### LINKED ISSUE #27: Add --name to "docker run" commands in workflow-helper script
This will make reading logs easier, as then one can just do `docker logs fluent-bit` if needed and should prevent multiple containers to run simultaneously, if startup script is re-executed.

parauliya:
## Description

Container name was not provided for `tink-worker` container.

## Why is this needed

Fixes: #27 

## How Has This Been Tested?
I have varified the change by creating a `tink-worker` through vagrant.

### DIFF
--- a/apps/workflow-helper.sh
+++ b/apps/workflow-helper.sh
@@ -58,7 +58,7 @@ rm -f /sbin/mdev
 
 mkdir /worker
 
-docker run --privileged -t \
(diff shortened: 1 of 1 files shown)


thebsdbox (approved):
This should make logs much easier to determine.

PR #79, opened 2020-07-31
Integrate boot file make into top level

mmlb:
## Description

Lift the rules from installer/alpine/Makefile into rules.mk.j2 so that the top-level make can handle building the alpine files itself. Also cleanup/fix the Dockerfile & build.sh files that are responsible for building the Alpine boot files.

## Why is this needed

Extracted out of #72 for better reviewability.

## How Has This Been Tested?
Builds kernel, initramfs, and modloop files as needed.


installer/alpine/Dockerfile
```diff
@@ -1,100 +1,101 @@
...
+ENV KERNEL 4.14.55
+ENV MAKEFLAGS -j$(($(nproc) * 2))
+# make sure PKGREL is +1 what is in APKGBUILD
+ENV PKGREL 1
+
+RUN true && \
+    # Setup self-signed keys to sign the built packages \
+    abuild-keygen -a -i -n && \
+    # Fetch alpine package tree \
+    curl -fL https://github.com/alpinelinux/aports/archive/$COMMIT.tar.gz | tar -zxf - && \
```
DavidHwu, line 55: do we care about failure scenarios ?

mmlb: More than handling via pipefail? https://github.com/tinkerbell/osie/pull/79/files#diff-b67911656ef5d18c4ae36cb6741b7965R2

rules.mk.j2
```diff
@@ -220,23 +220,27 @@ else
 	ci/vm.sh network_test | tee build/$@.log >/dev/$T
 endif
 
+ifneq ($(CI),drone)
+build/osie-aarch64.tar.gz: /proc/sys/fs/binfmt_misc/qemu-aarch64
+endif
 build/osie-aarch64.tar.gz: SED=/FROM/ s|.*|FROM multiarch/ubuntu-debootstrap:arm64-xenial|
-build/osie-x86_64.tar.gz: SED=
+build/osie-x86_64.tar.gz:  SED=
```
splaspood, line 227: extra space, no?

rules.mk.j2
```diff
@@ -220,23 +220,27 @@ else
...
-	$(Q)docker save osie:$* > $@.tmp
-	$(Q)mv $@.tmp $@
+	$(Q)docker save osie:$* > $@
 
+ifneq ($(CI),drone)
+build/osie-runner-aarch64.tar.gz: /proc/sys/fs/binfmt_misc/qemu-aarch64
+endif
 build/osie-runner-aarch64.tar.gz: SED=/FROM/ s|.*|FROM multiarch/alpine:arm64-v3.7|
-build/osie-runner-x86_64.tar.gz: SED=
+build/osie-runner-x86_64.tar.gz:  SED=
```
splaspood, line 239: another extra space

splaspood (changes requested):

installer/alpine/Dockerfile
DavidHwu: My question was geared around what error handling strategy we use... as I see this in docker file I understand now.  Withdrawing.

rules.mk.j2
mmlb: lol, @mikemrm said the same thing. I did it on purpose to line up the SEDs in the Makefile, it was annoying me to be unaligned.

rules.mk.j2
mmlb: ditto

DavidHwu (dismissed):

splaspood (dismissed):

mmlb:
@splaspood PTAL at the last 2 commits. These setup a way to skip building the Alpine files and adds some updated docs to the README.

splaspood (dismissed):

mmlb:
Trying to figure out why drone/make is still trying to rebuild the alpine assets.

mmlb:
Finally got the "skip building alpine assets" stuff correct.

@splaspood @DavidHwu one more looksie please.

apps/runner.sh
```diff
@@ -86,9 +86,14 @@ phone_home_url=$(jq -r .phone_home_url "$metadata")
 trap fail EXIT
 
 service docker start
-until docker info; do
+i=0
+# shellcheck disable=SC2169
+until docker info >/dev/null || [[ $i -gt 5 ]]; do
```
splaspood, line 91: I feel like this 'wait for docker to be ready' code that's duplicated should be pulled out into a shared function...  but not strongly enough to block the review on.

apps/workflow-helper.sh
```diff
@@ -42,11 +44,14 @@ mkdir -p /etc/docker/ /etc/docker/certs.d/ "/etc/docker/certs.d/$docker_registry
 cp ca.pem "/etc/docker/certs.d/$docker_registry/ca.crt"
 
 service docker start
-
-until docker info; do
-	echo 'Waiting for docker to respond...'
+i=0
+# shellcheck disable=SC2169
+until docker info >/dev/null || [[ $i -gt 5 ]]; do
```
splaspood, line 49: see above

apps/osie-installer.sh
```diff
@@ -135,9 +133,15 @@ cat <<-EOF | grep -v '^\s*$'
 	${packet_base_url:+packet_base_url: $packet_base_url}
 EOF
 
-until docker info; do
+service docker start
+i=0
+# shellcheck disable=SC2169
+until docker info >/dev/null || [[ $i -gt 5 ]]; do
```
splaspood, line 139: see below

installer/alpine/Dockerfile
```diff
@@ -1,100 +1,98 @@
...
-
-RUN sed -i 's/silentoldconfig/olddefconfig/g' APKBUILD && \
-    echo 'CONFIG_KEXEC=y' >> config-vanilla.x86_64 && \
+ENV COMMIT bb52b0d3f60cd75197ef218362f193378dfd869d
+ENV FLAVOR vanilla
+ENV KERNEL 4.14.55
+# make sure PKGREL is +1 what is in APKGBUILD
+ENV PKGREL 1
+
+RUN true && \
```
splaspood, line 51: true && ?

shell.nix
```diff
@@ -30,14 +30,15 @@ let
...
+    cpio
     curl
     docker
     git
     git-lfs
     gnumake
     gnused
     libarchive
-    minio
+    minio-client
```
splaspood, line 41: Is this even used anymore?

splaspood (changes requested):

apps/osie-installer.sh
splaspood: doh, I figured it put these comments in order of the code, not in the order that I left them...  anyway.. you get the idea ;)

apps/runner.sh
mmlb: Would agree but this other file would also have to be fetched from /init (really we should just embed them into the initramfs) which I kind of don't really want to do.

apps/workflow-helper.sh
mmlb: ditto

apps/osie-installer.sh
mmlb: heh

installer/alpine/Dockerfile
mmlb: I want to keep the comment above the command but docker won't let me start a `RUN` with a comment so I just used `true` for that.

shell.nix
mmlb: actually yes, we use it to upload to s3. We could use a different tool, but there isn't really a need to change it in this PR.

splaspood: ahh, ok.

splaspood (approved):

PR #86, opened 2020-08-10
Update versioning scheme

mmlb:
## Description

Updates osie version numbering scheme to match those of other services built in drone.

## Why is this needed

Aligns with other version numbering scheme we use. Also helps to track down builds in drone.packet.net history as there's no pagination support there.

## How Has This Been Tested?

Tested this locally and does what we need.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,4 +1,7 @@
 ---
+osie-test-env: &osie-test-env
+  image: osie-test-env:${DRONE_BUILD_NUMBER}-${DRONE_COMMIT_SHA}
+
 clone:
   git:
     image: quay.io/packet/drone-git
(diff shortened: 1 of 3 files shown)


gianarb:
@mmlb is this PR still valid? #118 is this a duplicate?

mmlb:
@gianarb it's stalled for the moment, and not directly related to #118. This is about changing the version numbers we use for osie, while #118 is about embedding the version info *into* osie so it can become self-referential.

tstromberg:
PR looks good to me - is anything else required other than updating the merge conflict?

mmlb:
> PR looks good to me - is anything else required other than updating the merge conflict?

Rebased on master, and realized my initial attempt at changing the value `v` was no where near as wide reaching as needed. Waiting for aarch64 test to finish, but pretty sure it'll pass. So good to merge.

mmlb:
Testing @ScottGarman

ScottGarman (approved):

PR #87, opened 2020-08-12
move early phone home to after install completes

mikemrm:
custom images are phoning home too early and causing switch configs
to be modified too soon breaking network connectivity during the
preinstall. Moving this to after the install completes delays any
post network configuration from happening allowing the install to
complete.

Signed-off-by: Mike Mason <mason@packet.com>

### DIFF
--- a/docker/scripts/osie.sh
+++ b/docker/scripts/osie.sh
@@ -97,11 +97,6 @@ else
 	custom_image=true
 fi
 
-# Phone home to tink NOW if non-packet custom image is used
-if [ "$early_phone" -eq 1 ]; then
-	phone_home "${tinkerbell}" '{"instance_id":"'"$(jq -r .id "$metadata")"'"}'
-fi
-
 target="/mnt/target"
 cprconfig=/tmp/config.cpr
 cprout=/statedir/cpr.json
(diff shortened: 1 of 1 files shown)


mmlb:
hmm thats not a good idea, we do want to phone-home early actually before we try to fetch any kind of custom image in case its a bad image/url we don't lose a host while it goes through deprov.

mikemrm:
i can see that reasoning. if that's the case, then we need to either, 1. re-dhcp so we're using the assigned ip addresses so when the api does post active cleanup tasks, we don't lose connectivity. or 2. change the cleanup tasks to trigger on a different event. from what i had understood, historically, we phoned home at the end of the install. but i must have been informed incorrectly, as i've been unable to confirm that.

ScottGarman:
Thanks @mmlb  for pointing out the need for the early phone_home. I've created another PR that takes Mike's suggested approach number 1 to fix this, and it has passed my testing: https://github.com/tinkerbell/osie/pull/89

If that PR is accepted, this one should be closed.

mmlb:
@mikemrm we phone-home from the instance after it boots, if no non-packet-approved things are done. custom image is one of those "sure, but if you break it you buy it" or w/e.

PR #89, opened 2020-08-13
Always re-DHCP during custom OS image installs

ScottGarman:
Without this, the early phone_home will mark the instance as activated,
and narwhal will remove the hardware IP address after a variable period
of time (typically less than a minute) at which point the provision will
fail because it can't download the image or phone_home to tinkerbell.

### DIFF
--- a/docker/scripts/osie.sh
+++ b/docker/scripts/osie.sh
@@ -97,8 +97,12 @@ else
 	custom_image=true
 fi
 
-# Phone home to tink NOW if non-packet custom image is used
+# Phone home to tink NOW if non-packet custom image is used. We don't do this
+# later in case the custom OS image or url is bad, to ensure instance will be
+# preserved for the user to troubleshoot.
 if [ "$early_phone" -eq 1 ]; then
+	# Re-DHCP so we obtain an IP that will last beyond the early phone_home
+	reacquire_dhcp "$(ip_choose_if)"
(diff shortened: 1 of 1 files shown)


ScottGarman:
I've successfully done 5 custom os provisions in a row using an Ubuntu 16.04 image Dustin originally reported this issue when trying to test. In the timelines for these instances, the "Provisioning complete" message still shows up early in the timeline, and my understanding is that this is expected due to the early phone_home that causes the API to be marked as active. But my tests now show that networking continues to be enabled through the rest of the provisioning process, so I think this is a complete fix for the issue.

docker/scripts/osie.sh
```diff
@@ -95,9 +95,14 @@ else
 	fi
 	early_phone=1
 	custom_image=true
+
+	# Re-DHCP so we obtain an IP that will last beyond the early phone_home
+	reacquire_dhcp "$(ip_choose_if)"
```
mmlb: I think this should be down below when we act on the `early_phone -eq 1` case. Lets keep all the action bits together.

mmlb (dismissed):

docker/scripts/osie.sh
ScottGarman: That's reasonable. I've updated the PR with that change.

mmlb (approved):

mikemrm (approved):

PR #90, opened 2020-08-16
API provisioning log event improvements

ScottGarman:
A few minor tweaks to when certain phone_home log event messages are
sent, and the addition of a new event to declare when the OS image has
been retrieved. The code comments have been updated to reflect the
current state of the event messages we're using for improved
readability.

### DIFF
--- a/docker/scripts/osie.sh
+++ b/docker/scripts/osie.sh
@@ -65,6 +65,9 @@ BASEURL=${BASEURL:-http://install.$facility.packet.net/misc/osie/current}
 # location, we should not trample it
 mirror=${mirror:-http://mirror.$facility.packet.net}
 
+## Tell the API we're connected to the magic install system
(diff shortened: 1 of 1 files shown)


mmlb:
@ScottGarman I converted to a Draft PR and dropped the WIP. This way we're guarded against merging it with the WIP in the title by mistake.

mmlb (commented):
lgtm

ScottGarman:
I've tested this today and it does what I intend for it to do. I do wish there was a way to issue the 104 provisioning event sooner (Connected to magic install system), as there is almost no time delay between that event firing and the new OS image retrieval event. But in the case where the OS image retrieval fails, this will provide useful information for debugging.
I'll be on PTO Tuesday but as soon as I'm back I'll push to have this merged and deployed, and then I can go about updating the OSV metadata for various OSes to include the new event number.

mmlb (approved):

PR #91, opened 2020-08-17
Eng 8289 update packetnetworking symlink fix

dlaube:
## Description

Update the packet-networking submodule

## Why is this needed

This is to add a bugfix for the symlink issue around resolv.conf

## How Has This Been Tested?
Testing via OSIE on real hardware in a single facility

None.


mmlb (dismissed):

mmlb:
@dlaube needs dco signoff, also can you squash all the commits into just one. You can do `git rebase --signoff`

mikemrm (approved):

PR #94, opened 2020-08-21
if ip_addresses is missing return an empty array

mikemrm:
in some instances this may be empty if we're wanting the host to
simply wipe the disks and restart. This change keeps a KeyError
exception from being raised.

### DIFF
--- a/osie-runner/handlers.py
+++ b/osie-runner/handlers.py
@@ -236,7 +236,7 @@ def cacher_to_metadata(j, tinkerbell):
         "hostname": instance["hostname"],
         "id": instance["id"],
         "network": {
-            "addresses": instance["ip_addresses"],
+            "addresses": instance.get("ip_addresses", []),
             "bonding": {"mode": j["bonding_mode"]},
             "interfaces": [
                 {"bond": p["data"]["bond"], "mac": p["data"]["mac"], "name": p["name"]}


osie-runner/run.py
```diff
@@ -81,6 +82,11 @@ def connect_hegel(facility):
     tinkerbell = parse.urlparse(util.value_from_kopt(cmdline_content, "tinkerbell"))
     facility = util.value_from_kopt(cmdline_content, "facility")
 
+handler = logging.handlers.SysLogHandler(
+    address=(f"tinkerbell.{facility}.packet.net", 514)
+)
+log.addHandler(handler)
+
```
mikemrm: One thing to take into consideration with this, is this might cause an exception to be raised if the SysLogHandler is unable to connect to the syslog server. A custom handler might need to wrap the sysloghandler in order to capture and dismiss failed remote syslog write attempts.

mikemrm: splitting this out into a separate pr

osie-runner/handlers.py
```diff
@@ -236,7 +236,7 @@ def cacher_to_metadata(j, tinkerbell):
         "hostname": instance["hostname"],
         "id": instance["id"],
         "network": {
-            "addresses": instance["ip_addresses"],
+            "addresses": instance.get("ip_addresses", []),
```
mmlb, line 239: I'm now very cautious about this with the data api is ultimately responsible for. I think we should walk the safe side and do
```python
addr = instance.get("ip_addresses")
if not addr:
    addr = []
```

mikemrm: I believe the main problem, is hegel now has a struct for instance to be parsed into rather than the interface it got tossed into before. In order to be backwards compatible, we had to put `omitempty` on all the fields inside the instance struct so the result for an empty instance is `instance: {}` however, this now breaks in this case where the api may be pushing an empty array of ips, but because we have omitempty, it's not including that attribute in the output.

Also a one line example could be `"addresses": instance.get("ip_addresses") or [],` which is the same thing

ScottGarman (approved):

osie-runner/handlers.py
mmlb: ah yeah I was wanting a one-liner and forgot about simple `or ...`

PR #96, opened 2020-08-24
pip install packet-hardware and packet-networking

mikemrm:
## Description

Switch to using packethost/packet-hardware and packet-networking repo instead of code being within project.

Also removed legacy network tests as those test are now handled within the packet-networking repo

## Why is this needed

This allows for easier development of the tools and is able to be used in more ways.


gianarb (dismissed):

mmlb:
@mikemrm PR title and comment needs updating

mmlb (commented):
PR title and body needs updating. Mergify will use those for the merge message so should be kept up2date.

mmlb (changes requested):
gah, I need some  coffee

mmlb (approved):

mmlb:
We could probably slim down the `ci/Dockerfile` image now a bit too. I know we don't need lshw in there anymore, but fine to leave for now.

PR #99, opened 2020-08-25
Log OSIE failures to tinkerbell using a bash trap function

ScottGarman:
Similar to what we do in osie-runner, this sets up a bash trap
function named autofail() which logs to tinkerbell an error message
based on the `autofail_reason` variable, which we update in various
sections of the script to reflect which code section the failure
occurred.

### DIFF
--- a/docker/scripts/osie.sh
+++ b/docker/scripts/osie.sh
@@ -55,6 +55,17 @@ if [[ $state == 'osie.internal.check-env' ]]; then
 	exit 0
 fi
 
+# On errors, run autofail() before exiting
+autofail_reason='unknown, just started osie'
+function autofail() {
+	# Passthrough for when the main script exits normally
+	# shellcheck disable=SC2181
+	(($? == 0)) && exit
+
+	puttink "${tinkerbell}" phone-home '{"type":"failure", "reason":"'"${autofail_reason}"'"}'
+}
+trap autofail EXIT
+
 OS=$os${tag:+:$tag}
 
 # if $BASEURL is not empty then the user specifically passed in the artifacts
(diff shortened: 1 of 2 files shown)


mikemrm (dismissed):
LGTM!

docker/scripts/osie.sh
```diff
@@ -55,6 +55,17 @@ if [[ $state == 'osie.internal.check-env' ]]; then
...
+# On errors, run autofail() before exiting
+autofail_reason='unknown, just started osie'
+function autofail() {
+	if [[ "${autofail_reason}" == "SUCCESS" ]]; then
+		return 0
+	fi
+	puttink "${tinkerbell}" phone-home '{"type":"failure", "reason":"'"${autofail_reason}"'"}'
+	print_stack_trace
+}
+trap autofail EXIT
```
mmlb, line 68: change this to `ERR` and you can do away with the `== "SUCCESS"` handling

docker/scripts/osie.sh
```diff
@@ -587,3 +609,4 @@ EOF
 fi
 
 chmod +x /statedir/cleanup.sh
+autofail_reason='SUCCESS'
```
mmlb: can go away with `trap ... ERR`

mmlb (changes requested):

ScottGarman:
Changing this to an ERR trap isn't working, I'll research this more tomorrow.

ScottGarman:
Manny and I discussed the trap on EXIT vs. ERR issue this morning, and I've implemented the solution we came to. This is ready for final review and merge now.

mmlb (approved):

PR #101, opened 2020-08-26
Eng 8520 nsx admin user

dlaube:
## Description

This makes the set password function more generic in order to support alternative superuser names (admin) in the case of VMware NSX.

## Why is this needed

This is required in order to keep the default superuser name for VMware NSX

Fixes: #

## How Has This Been Tested?
Testing custom OSIE release for provisions of ubuntu, virtuozo (vosie) and NSX.

None.

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/docker/scripts/osie.sh
+++ b/docker/scripts/osie.sh
@@ -573,7 +573,11 @@ fi
 
 autofail_reason='error setting root password'
 echo -e "${GREEN}#### Setting password${NC}"
-set_pw "root" "$pwhash" $target/etc/shadow
+pwuser="root"
+if [[ ${OS} =~ vmware_nsx_3_0_0 ]]; then
+	pwuser="admin"
+fi
+set_pw "$pwuser" "$pwhash" $target/etc/shadow
 
 # ensure unique dbus/systemd machine-id, will be based off of container_uuid aka instance_id
 autofail_reason='error setting machine-id'
--- a/docker/scripts/osie.sh
+++ b/docker/scripts/osie.sh
@@ -572,8 +572,8 @@ if [[ $pwhash == "preinstall" ]]; then
 fi
 
 autofail_reason='error setting root password'
-echo -e "${GREEN}#### Setting root password${NC}"
-set_root_pw "$pwhash" $target/etc/shadow
+echo -e "${GREEN}#### Setting password${NC}"
+set_pw "root" "$pwhash" $target/etc/shadow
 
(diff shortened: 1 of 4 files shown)


docker/scripts/osie.sh
```diff
@@ -542,8 +542,14 @@ if [[ $pwhash == "preinstall" ]]; then
...
-echo -e "${GREEN}#### Setting root password${NC}"
-set_root_pw "$pwhash" $target/etc/shadow
+echo -e "${GREEN}#### Setting password${NC}"
+if [[ ${OS} =~ vmware_nsx_3_0_0 ]] ; then
+	pwuser=admin
+	set_pw "$pwuser" "$pwhash" $target/etc/shadow
+else
+	pwuser=root
+	set_pw "$pwuser" "$pwhash" $target/etc/shadow
+fi
```
mmlb: All we need in the if state is pwuser and not the call to `set_pw`, lets do this instead:
```sh
pwuser=root
if [[ ${OS} =~ ^vmware_nsx ]]; then
    pwuser=admin
fi
set_pw pwuser "$pwhash" $target/etc/shadow
```

docker/scripts/osie.sh
```diff
@@ -542,8 +542,14 @@ if [[ $pwhash == "preinstall" ]]; then
...
-echo -e "${GREEN}#### Setting root password${NC}"
-set_root_pw "$pwhash" $target/etc/shadow
+echo -e "${GREEN}#### Setting password${NC}"
+if [[ ${OS} =~ vmware_nsx_3_0_0 ]]; then
+	pwuser="admin"
+	set_pw "$pwuser" "$pwhash" $target/etc/shadow
+else
+	pwuser="root"
+	set_pw "$pwuser" "$pwhash" $target/etc/shadow
+fi
```
mmlb: since only pwuser changes, lets only mess with that. Also, it should be ok to be a bit more generic and support all vmware_nsx values right? So:
```sh
pwuser=root
[[ ${OS} =~ ^vmware_nsx ]] && pwuser=admin

set_pw "$pwuser" "$pwhash" $target/etc/shadow 
```

mmlb (changes requested):

docker/scripts/osie.sh
dlaube: Yep it makes sense to only mess with that. Nope, other/older versions of NSX appear to use root for default.

mmlb (dismissed):
:shipit:

mmlb (approved):

PR #102, opened 2020-08-27
Gather and report CPLD version and more

dustinmiller:
## Description

Lets add racadm and ipmicfg and update packet-hardware.

## Why is this needed

Need these so we can gather the CPLD version and more from Dell and Supermicro equipment

## How Has This Been Tested?
Tested from within OSIE container and on several Ubuntu 16 machines
Corresponds with packethost/packet-hardware/pull/1

No break. Only fix. 🤞

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -118,6 +118,24 @@ RUN cd /tmp/osie && \
     install -m 755 /tmp/osie/ipmicfg /usr/bin/ipmicfg && \
     rm -rf /tmp/osie
 
+# RACADM
+COPY dchipm.ini /tmp/osie/
+RUN cd /tmp/osie && \
+    if [ $(uname -m) != 'aarch64' ]; then \
+      apt-get update && apt-get install -y alien && \
+      rpm --import http://linux.dell.com/repo/pgp_pubkeys/0x1285491434D8786F.asc && \
+      wget \
+        https://dl.dell.com/FOLDER05920767M/1/DellEMC-iDRACTools-Web-LX-9.4.0-3732_A00.tar.gz \
+        http://linux.dell.com/repo/community/openmanage/940/bionic/pool/main/s/srvadmin-omilcore/srvadmin-omilcore_9.4.0_amd64.deb && \
+      tar -xvf DellEMC-iDRACTools-Web-LX-9.4.0-3732_A00.tar.gz && \
+      alien -i iDRACTools/racadm/RHEL8/x86_64/*.rpm && \
+      dpkg -i *.deb && \
+      apt-get purge -y alien && \
+      apt-get autoremove -y && \
+      cp dchipm.ini /opt/dell/srvadmin/etc/srvadmin-hapi/ini/ ; \
+    fi && \
+    rm -rf /tmp/osie
+
 # URL=http://www.mellanox.com/downloads/firmware/mlxup
 # VERSION=4.6.0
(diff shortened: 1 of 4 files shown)


docker/Dockerfile
```diff
@@ -74,7 +74,7 @@ RUN apt-get update -y && \
     apt-get -qy clean && \
     rm -rf /var/lib/apt/lists/* /tmp/osie/
 
-ARG PACKET_HARDWARE_COMMIT=4702a57d31a86c12ae6b016a88edcc15ee64a42d
+ARG PACKET_HARDWARE_COMMIT=b4e85b9fe707573bcc54954c4f85e620f5c8753f
```
mmlb, line 77: Why? (please answer in commit message :D )

docker/Dockerfile
```diff
@@ -112,6 +112,12 @@ RUN cd /tmp/osie && \
     cd && \
     rm -r /tmp/osie
 
+# IPMICFG
+COPY lfs/ipmicfg /tmp/osie/
```
mmlb, line 116: It'd be good to have in the commit message the url and version of this file

docker/Dockerfile
```diff
@@ -118,6 +118,22 @@ RUN cd /tmp/osie && \
     install -m 755 /tmp/osie/ipmicfg /usr/bin/ipmicfg && \
     rm -rf /tmp/osie
 
+# RACADM
+COPY dchipm.ini /tmp/osie/
+RUN cd /tmp/osie && \
+    if [ $(uname -m) != 'aarch64' ]; then \
+      apt-get update && apt-get install -y alien && \
```
mmlb, line 125: should `purge alien` when done

mmlb (changes requested):

docker/Dockerfile
dustinmiller: Yup 👍

docker/Dockerfile
dustinmiller: Added 🙂

docker/Dockerfile
dustinmiller: Indeed, purging 🧹

docker/Dockerfile
```diff
@@ -118,6 +118,23 @@ RUN cd /tmp/osie && \
...
+      rpm --import http://linux.dell.com/repo/pgp_pubkeys/0x1285491434D8786F.asc && \
+      wget \
+        https://dl.dell.com/FOLDER05920767M/1/DellEMC-iDRACTools-Web-LX-9.4.0-3732_A00.tar.gz \
+        http://linux.dell.com/repo/community/openmanage/940/bionic/pool/main/s/srvadmin-omilcore/srvadmin-omilcore_9.4.0_amd64.deb && \
+      tar -xvf DellEMC-iDRACTools-Web-LX-9.4.0-3732_A00.tar.gz && \
+      alien -i iDRACTools/racadm/RHEL8/x86_64/*.rpm && \
+      dpkg -i *.deb && \
+      apt-get purge -y alien && \
```
mmlb, line 133: I think we also need `apt-get autoremove -y`

mmlb (changes requested):

docker/Dockerfile
dustinmiller: Good call, added ✔️

mmlb (approved):

PR #105, opened 2020-08-28
Add an .editorconfig file for this repo

ScottGarman:
Specifies tab indent style for bash scripts and space indent style for python scripts.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,5 @@
+[*.sh]
+indent_style = tab
+
+[*.py]
+indent_style = space
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -1,4 +1,3 @@
-# vim: set expandtab shiftwidth=4 softtabstop=4:
 FROM alpine:3.12
 
 RUN apk add --no-cache --no-progress --update --upgrade \


ScottGarman:
I'm happy to add additional modeline options if desired.

mmlb:
I think it might be better to just setup an [editorconfig](https://editorconfig.org/) file instead this way it works for all editors.

mmlb:
@ScottGarman thoughts on my editorconfig comment?

mmlb (commented):
maybe better off as an editorconfig file instead?

ScottGarman:
@mmlb I've finally gotten around to updating this PR to use editorconfig. Just specifying the indent style, which is the minimum needed to restore my sanity again.

mmlb:
@ScottGarman can I get you to remove the modelines in all the files present too? Don't want 2 sources for the settings.

mmlb (dismissed):

ScottGarman:
@mmlb it turned out there was only one file we had (ci/Dockerfile) that had vim modelines set that could be trimmed out. There are several others that come in from the shunit submodule that aren't our code that I'm leaving out.

mmlb (approved):

PR #108, opened 2020-08-28
DO NOT MERGE: Do not automatically fail provisions when autofail() is triggered

ScottGarman:
We seem to be having instances that are failing with an autofail
message, but the provision continues and the system appears to complete
the provisioning process anyway. Until we have a clearer idea of how
this is happening, do not mark provisions as failed.

### DIFF
--- a/docker/scripts/osie.sh
+++ b/docker/scripts/osie.sh
@@ -63,7 +63,7 @@ function autofail() {
 	(($? == 0)) && exit
 
 	print_stack_trace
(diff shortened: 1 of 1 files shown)


mikemrm (approved):

jgavinray (approved):

ScottGarman:
I'm not going to merge this for time time being, I'd rather troubleshoot and get to the bottom of why this is occurring. In the meantime I've rolled back OSIE in production to before the autofail() function was added.

PR #118, opened 2020-09-10
Track OSIE version information

ScottGarman:
Normally I'd create a file called VERSION that would get packaged up and could be read from, but since we're generating multiple containers, ensuring access to this file gets a bit complicated. So I thought it would be easier to template some variables in specific OSIE scripts that could be substituted at build time.

Consider this an RFC pull request and if we can get consensus on the technical approach for this, there are several more areas where we can make use of this version information - e.g, sending it to hegel on initial connection.

### DIFF
--- a/apps/runner.sh
+++ b/apps/runner.sh
@@ -60,6 +60,7 @@ ______          _        _                _
 \_|  \__,_|\___|_|\_\___|\__(_)_| |_|\___|\__|
 ===============================================
 ##          Task Runner Environment          ##
+## OSIE Version: ${OSIE_VERSION} (${OSIE_BRANCH})
 EOF
 
 facility=$(sed -nr 's|.*\bfacility=(\S+).*|\1|p' /proc/cmdline)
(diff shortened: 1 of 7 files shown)


mikemrm:
What do you think about just including it as args in the dockerfile and creating a .env file during build time that can just get sourced?

ScottGarman:
@mikemrm That's an interesting idea. I think I like it. That way those variables could be referenced in any of our scripts and not have to be set manually in each one. I'll hold off on going that route until @mmlb has chimed in with feedback too.

.drone.yml
```diff
@@ -55,6 +55,7 @@ pipeline:
     group: ci1
     image: osie-test-env
     commands:
+      - bash set_version.sh
```
mmlb: Hmm this should be in the Makefile instead.

set_version.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+VERSION=`git rev-parse HEAD`
+BRANCH=`git symbolic-ref --short HEAD`
```
mmlb: We should add the drone build number too

mmlb (approved):

PR #119, opened 2020-09-12
Fix file perms when generating initramfs

ScottGarman:
On Ubuntu 18.04 and 20.04 I discovered that building OSIE locally would
result in a failure because the initramfs would get copied out of the
container as owned by root with user-only read permissions. This updates
the file permissions to ensure the file can be read by the build user.


mikemrm (approved):

ScottGarman:
I snuck this commit into my OSIE version info PR which has been merged, so closing this.

PR #120, opened 2020-09-14
Stop building git sub-packages to speed up CI builds

ScottGarman:
We were previously building all of the git sub-packages that we have no
use for, and it was eating up a ton of build time (especially for
aarch64 builds).

### DIFF
--- a/docker/build-git-openssl.sh
+++ b/docker/build-git-openssl.sh
@@ -32,9 +32,20 @@ apt-get install -y libcurl4-openssl-dev
 apt-get source git
 (
 	cd git*
+	# Change the libcurl4-gnutls-dev dep to libcurl4-openssl-dev
 	sed -i debian/control \
 		-e 's/libcurl4-gnutls-dev/libcurl4-openssl-dev/' \
 		-e '/TEST\s*=\s*test/d' ./debian/rules
+	# Strip out all git-xyz packages from the control file to speed up builds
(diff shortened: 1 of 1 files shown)


ScottGarman:
This appears to have shaved off at least 15m from the aarch64 OSIE build, which is the long pole in our CI pipeline.

mikemrm (approved):

docker/build-git-openssl.sh
```diff
@@ -32,16 +32,28 @@ apt-get install -y libcurl4-openssl-dev
 apt-get source git
 (
 	cd git*
+	# Change the libcurl4-gnutls-dev dep to libcurl4-openssl-dev
 	sed -i debian/control \
 		-e 's/libcurl4-gnutls-dev/libcurl4-openssl-dev/' \
 		-e '/TEST\s*=\s*test/d' ./debian/rules
+	# Strip out all git-xyz packages from the control file to speed up builds
+	sed -i debian/control \
+		-e '/^Package: git-man/,$d' ./debian/control
```
mmlb, line 41: @ScottGarman this doesn't match the comment?

ScottGarman: Perhaps I could have explained it better - the sed basically removes everything in the control file starting with the Package definition of git-man through the end of the file. I can improve this comment further if you think that would clarify things.

PR #126, opened 2020-09-16
Fix Custom OS image installs for our sites that don't support ping network checks

ScottGarman:
This PR includes two fixes. The first is a general fix to properly detect which ethernet interface to use when re-dhcp'ing. The second eliminates ping-based checks for custom OS image installs and in our github-mirror check function.

For custom OS images, we already re-dhcp, so if there a bad git url or networking issue, allow the provision to fail during the git clone. For the github-mirror check, add a re-dhcp if the initial clone of the lfs-test repo fails.


mikemrm (approved):
Awesome!

PR #127, opened 2020-09-16
Use pre-built binary deb packages for git

ScottGarman:
We use custom git packages that have been built with openssl instead of
gnutls. Rather than rebuild the packages on every CI run, start using
prebuilt ones, which were built using the build-git-openssl.sh script.
That script has been updated and includes documentation in the file
header.

### DIFF
--- a/docker/build-git-openssl.sh
+++ b/docker/build-git-openssl.sh
@@ -1,9 +1,12 @@
 #!/usr/bin/env bash
+#
+# Run this script on x86 and aarch Ubuntu 16.04 hosts to generate binary git
+# packages built with openssl instead of gnutls. Then copy the main git deb
+# files into the docker/lfs directory of this repo, and rename them to end
+# with 'x86_64.deb' or 'aarch64.deb'.
 
 set -euxo nounset
 
-d=$(mktemp -d)
-cd "$d"
 apt-get install -y software-properties-common
 # The version of git that comes with Xenial (2.7.4) doesn't support shallow
 # clones from our github-mirror server, so use the latest officially supported
@@ -36,13 +39,13 @@ apt-get source git
 	sed -i debian/control \
 		-e 's/libcurl4-gnutls-dev/libcurl4-openssl-dev/' \
 		-e '/TEST\s*=\s*test/d' ./debian/rules
-	# Strip out all git-xyz packages from the control file to speed up builds
+	# Strip out all git-xyz subpackages from the control file to speed up builds
 	sed -i debian/control \
 		-e '/^Package: git-man/,$d' ./debian/control
-	# Remove git-man from git's Depends (fragile, based on current control file)
+	# Remove git-man from git's Depends section
 	sed -i debian/control \
 		-e '/git-man/d' ./debian/control
-	# Remove the trailing comma, now that this Depends is one line
+	# Now git's Depends section is one line, so strip off the trailing comma
 	sed -i debian/control \
 		-e 's/^\(Depends:.*liberror-perl\),$/\1/' ./debian/control
 
@@ -53,7 +56,8 @@ apt-get source git
 	cat >>debian/changelog.tmp <<-EOF
 		
 		  * rebuild with openssl instead of gnutls
-		  * remove subpackages and Depends on git-man in debian/control file
+		  * remove subpackages from debian/control, only build the main git package
+		  * remove the Depends on git-man from the git main git package
 		
 		 -- OSIE Builder <osie-builder@localhost>  $(date +'%a, %d %b %Y %T %z')
 		
@@ -64,8 +68,3 @@ apt-get source git
 
 	dpkg-buildpackage -rfakeroot -b -j"$(nproc)"
 )
-mv ./*.deb /tmp/osie
-apt-get purge -y devscripts equivs libcurl4-openssl-dev software-properties-common
-apt-get autoremove -y
-cd
-rm -rf "$d"
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -32,11 +32,10 @@ RUN apt-get update -y && \
     apt-get -qy clean && \
     rm -rf /var/lib/apt/lists/* /tmp/osie
 
-# build openssl'd git, done here so we can keep it cached as long as possible
-COPY build-git-openssl.sh /tmp/osie/
+# install custom git package built with openssl instead of gnutls
+COPY lfs/git_2.28.0-1packethost1*.deb /tmp/osie/
 RUN apt-get update -y && \
-    /tmp/osie/build-git-openssl.sh && \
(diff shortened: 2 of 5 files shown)


ScottGarman:
This shaves about 12m from the aarch64 build time, and gets the build in well under 45m total now. In exchange for this, we need to store ~ 13 MB in git-lfs for the binary deb packages.

ScottGarman:
I've tested this with around 30 provisions across our facilities, and it's working well. Ready for final review & merge.

displague:
I'm naively looking at this for the first time. There are a lot of things done in this Dockerfile (before any of your changes) without persisting the justification, as such I wonder if these measures are still needed. (I'm not providing actual review of your changes here, I'm just trying to understand why we aren't approaching this in a different way.)

* Why is this using a custom build of git at all? Is git only used to access github.com? Git (that ships with ubuntu:xenial) doesn't appear to have any problems fetching from Github.  If the problem is with gnutls handshakes, there appear to be some alternatives to using a custom git, like set the sslVersion and algos through .gitconfig or environment variables: https://git-scm.com/docs/git-config#Documentation/git-config.txt-httpsslVersion

  If that's not sufficient, must xenial be used as the build environment?
* The next thing that jumps out is that this single image is built up with layer after layer of build tooling. With multistaged builds https://docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds these builds could be better cached and run in parallel. A change to one of the scripts would only trigger a (cached) rebuild of the build stage that COPYs that script and any other stages that COPY artifacts from those build stage.  There are volume and proxy strategies that could have the multistaged builds sharing deb downloads, but I think we may find these are not as necessary.

   The final image would copy the artifacts from the simpler build stages without the noise of the build stages.

* `dpkg --unpack ./file.deb && apt-get install` stanzas could become `apt install ./file.deb` (postinst manipulating occurences withstanding)

* mlxup has newer versions available at https://www.mellanox.com/support/firmware/mlxup-mft, whose URLs could be placed directly in the `COPY` commands (http://www.mellanox.com/downloads/firmware/mlxup/4.15.0/SFX/linux_x64/mlxup) (this could take advantage of arch variables and ARGS in the Dockerfile to set the version and arch of the mlxup file).

mmlb:
>     * Why is this using a custom build of git at all? Is git only used to access github.com? Git (that ships with ubuntu:xenial) doesn't appear to have any problems fetching from Github.  If the problem is with gnutls handshakes, there appear to be some alternatives to using a custom git, like set the sslVersion and algos through .gitconfig or environment variables: [git-scm.com/docs/git-config#Documentation/git-config.txt-httpsslVersion](https://git-scm.com/docs/git-config#Documentation/git-config.txt-httpsslVersion)

We had multiple intermittent provision failures seemingly due to GnuTLS bugs. The error messages were
> fatal: unable to access 'https://images.packet.net/packethost/packet-images.git/': gnutls_handshake() failed: Error in the pull function.

We suspected it had to do with some Github LB swaps or something. Never really got a good chance to reproduce for extended periods in order to debug better. The internets suggested swapping curl's backend from GnuTLS to openssl (should be able to search and find many cases of this) to fix, we did so and confirmed the fix. At this point I'd hesitate moving back to GnuTLS even if the bug was fixed (don't know if that's the case) as openssl seems to be healthy again and I'm seeing distros that jumped ship to libressl moving back anyway.

>       If that's not sufficient, must xenial be used as the build environment?

Build environment or OSIE `FROM`? I don't even want to think about running non-Xenial packages on Xenial. If the latter I'm very much up for upgrading to 20.04. I tried out 18.04 not long after it was released and ran into CI or prov issues, but did not have the time to track down/fix. Would welcome a PR updating to 20.04. Looks like we don't have an issue for that already (I thought did), will create one. (edit: #128) 

>     * The next thing that jumps out is that this single image is built up with layer after layer of build tooling. With multistaged builds [docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds](https://docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds) these builds could be better cached and run in parallel. A change to one of the scripts would only trigger a (cached) rebuild of the build stage that COPYs that script and any other stages that COPY artifacts from those build stage.  There are volume and proxy strategies that could have the multistaged builds sharing deb downloads, but I think we may find these are not as necessary.
>       The final image would copy the artifacts from the simpler build stages without the noise of the build stages.

For some reason I thought that busted caches from previous stages would also affect later ones, but I just ran a quick test and that was not the case (maybe it used to be?). This sounds like a great idea. PR welcome, I'll add an issue for it. (edit: #129)

>     * `dpkg --unpack ./file.deb && apt-get install` stanzas could become `apt install ./file.deb` (postinst manipulating occurences withstanding)

Not sure what you mean by postinst manipulating... but worth a try @ScottGarman.

>     * mlxup has newer versions available at [mellanox.com/support/firmware/mlxup-mft](https://www.mellanox.com/support/firmware/mlxup-mft), whose URLs could be placed directly in the `COPY` commands (http://www.mellanox.com/downloads/firmware/mlxup/4.15.0/SFX/linux_x64/mlxup) (this could take advantage of arch variables and ARGS in the Dockerfile to set the version and arch of the mlxup file).

This sounds fine too. I generally hesitate to to reach out to vendor websites as files tend to disappear post acquisition or rebrands or ... Looks like Mellanox/Nvidia are doing it right, atm, so I'm not against it. But I'm not sure what the benefit of changing would be?

docker/build-git-openssl.sh
```diff
@@ -1,9 +1,12 @@
 #!/usr/bin/env bash
+#
+# Run this script on x86 and aarch Ubuntu 16.04 hosts to generate binary git
+# packages built with openssl instead of gnutls. Then copy the main git deb
+# files into the docker/lfs directory of this repo, and rename them to end
+# with 'x86_64.deb' or 'aarch64.deb'.
```
mmlb, line 6: Doing the multistage build I think makes more sense I think. If not it would be nice to have a `rebuild.sh` or similar that does all these steps but run in a docker container, like:
```bash
case $arch in
x86_64) image=ubuntu:16.04;;
aarch64) image=the-ubuntu-qemu-image;;
*) error
esac

docker run --rm -ti -v $PWD:$PWD -w $PWD $image ./rebuild.sh
```

jacobweinstock:
Hey @ScottGarman, this has had the `do-not-merge` label for quite some time. Do we want to close this and re-open it when necessary?

ScottGarman:
Yeah, this was an aborted effort from a while back. I think this will be unnecessary once we move to a newer Ubuntu for osie anyway. Closing.

dustinmiller:
> Yeah, this was an aborted effort from a while back. I think this will be unnecessary once we move to a newer Ubuntu for osie anyway. Closing.

Also we're not building custom git anymore per https://github.com/tinkerbell/osie/pull/201

PR #132, opened 2020-09-18
Check for verbose_logging in userdata to enable xtrace

ScottGarman:
If '#verbose_logging=true' is found in userdata, enable xtrace for verbose console logging.

### DIFF
--- a/docker/scripts/osie.sh
+++ b/docker/scripts/osie.sh
@@ -55,6 +55,12 @@ if [[ $state == 'osie.internal.check-env' ]]; then
 	exit 0
 fi
 
+verbose_logging=$(sed -nr 's|.*\bverbose_logging=(\S+).*|\1|p' "$userdata")
+if [[ "${verbose_logging}" == true ]]; then
+	echo -e "${GREEN}#### Enabling Verbose OSIE Logging${NC}"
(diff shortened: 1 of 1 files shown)


ScottGarman:
I've got some extra noisy debug output to test that this feature is working as intended - I'll remove and rebase the PR once I've verified this works as expected.

docker/scripts/osie.sh
```diff
@@ -55,6 +55,15 @@ if [[ $state == 'osie.internal.check-env' ]]; then
 	exit 0
 fi
 
+echo -e "${GREEN}### CHECKING FOR VERBOSE LOGGING USERDATA ***${NC}"
+verbose_logging=$(sed -nr 's|.*\bverbose_logging=(\S+).*|\1|p' "$userdata")
```
mmlb, line 58: we should probably turn this into a function at some point, its used all over the place :D

mmlb (dismissed):

docker/scripts/osie.sh
ScottGarman: Agreed. I'll do that in a separate PR, though.

mmlb (approved):

PR #134, opened 2020-09-20
Refactor autofail feature to include tracking timeouts

ScottGarman:
The autofail() trap function in OSIE now tracks failures in two
contexts:

1. Non-zero exit codes will trigger autofail() and report an error
during the particular OSIE code stage
2. The current OSIE code stage is tracked in /statedir/autofail_stage so
it can bubble up to osie-runner, so we can reprort where hangs/timeouts
occur.

The $autofail_reason has been renamed to $autofail_stage, which is a
more accurate reflection of what it's used for.

### DIFF
--- a/apps/osie-installer.sh
+++ b/apps/osie-installer.sh
@@ -4,6 +4,17 @@
 
 reason='unknown'
 fail() {
+	if [ "$reason" = "docker exited with an error (osie-installer)" ]; then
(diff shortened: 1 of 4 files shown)


mmlb (commented):
lgtm

mmlb (approved):

PR #143, opened 2020-09-29
Enable autofail error logging for OSIE deprovision script

ScottGarman:
This enables the autofail() trap function and gives us error and timeout logging in the same way we've recently added for the osie.sh script.


mmlb (approved):

ScottGarman:
Going to spend a day testing this, will merge when ready.

PR #148, opened 2020-10-12
Add f3.medium GRUB templates

dustinmiller:
## Description

Adds in the GRUB templates for a new plan type, the f3.medium

## Why is this needed

Without these files, OS installations on this plan type do not succeed.

## How Has This Been Tested?
Barely, just following previous patterns.

No break, just fix. 🤞


mmlb (commented):
:+1:

ScottGarman (approved):

PR #150, opened 2020-10-19
Add Discovery OS mode

ScottGarman:
Discovery OS is a new mode of OSIE that offers various tools to
auto-detect and setup hardware during initial rack enrollment of a
server.
    
It's based on the packet/discover-metal container image we have
hosted on quay.io. During the OSIE build process we download this image
so it can be available during OSIE runs without having to retrieve it
via quay.

### DIFF
--- a/apps/discover.sh
+++ b/apps/discover.sh
@@ -1,0 +1,77 @@
+#!/bin/ash
+
+# shellcheck shell=dash
+
+arch=$(uname -m)
+packet_base_url=$(sed -nr 's|.*\bpacket_base_url=(\S+).*|\1|p' /proc/cmdline)
+facility=$(sed -nr 's|.*\bfacility=(\S+).*|\1|p' /proc/cmdline)
+tinkerbell=$(sed -nr 's|.*\btinkerbell=(\S+).*|\1|p' /proc/cmdline | sed -e 's|^http://||' -e 's|/.*||')
+phone_home_url="http://$tinkerbell/phone-home"
+
+discover_url=$(sed -nr 's|.*\bdiscover_url=(\S+).*|\1|p' /proc/cmdline)
+if [ -z "$discover_url" ]; then
+	discover_url="http://$tinkerbell/discover-os"
+fi
+
+reason='unknown'
+fail() {
+	curl -H 'Content-Type: application/json' \
+		-d '{"type":"failure", "reason":"'"$reason"'"}' \
+		"$phone_home_url"
+}
+
+# shellcheck disable=SC2039
+set -o errexit -o pipefail -o xtrace
+
+# Create rescue motd
+cat <<'EOF' >/etc/motd
+    ,------.               ,--.        ,--.                       
+    |  .---' ,---. ,--.,--.`--',--,--, `--',--.  ,--.             
+    |  `--, | .-. ||  ||  |,--.|      \,--. \  `'  /              
+    |  `---.' '-' |'  ''  '|  ||  ||  ||  | /  /.  \              
+    `------' `-|  | `----' `--'`--''--'`--''--'  '--'             
+,------.  ,--. `--'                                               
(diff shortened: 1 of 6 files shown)


apps/discover.sh
```diff
@@ -0,0 +1,77 @@
...
+
+arch=$(uname -m)
+packet_base_url=$(sed -nr 's|.*\bpacket_base_url=(\S+).*|\1|p' /proc/cmdline)
+facility=$(sed -nr 's|.*\bfacility=(\S+).*|\1|p' /proc/cmdline)
+tinkerbell=$(sed -nr 's|.*\btinkerbell=(\S+).*|\1|p' /proc/cmdline | sed -e 's|^http://||' -e 's|/.*||')
+phone_home_url="http://$tinkerbell/phone-home"
+
+discover_address=$(sed -nr 's|.*\bdiscover_address=(\S+).*|\1|p' /proc/cmdline)
+if [ -z "$discover_address" ]; then
+	discover_address="http://tinkerbell.$facility.packet.net"
```
mikemrm: Perhaps we could use the `$tinkerbell` variable instead?

mikemrm: Also, looking at discover-metal, it seems this should be the exact url to send to. So we might actually want to change this to something like `discover_url` or something along those lines. And with that perhaps it should be `/discover-metal` or `/discover-os` or something along those lines. Thoughts?

installer/alpine/init-aarch64
```diff
@@ -312,6 +315,11 @@ rescue)
 esac
 eend
 
+discover_address="tinkerbell.$KOPT_facility.packet.net"
+if [ "$packet_action" == "discover" ] && [ -n "$KOPT_discover_address" ]; then
+  discover_address=$KOPT_discover_address
+fi
```
mikemrm: Not sure this is necessary since we're grepping it out of /proc/cmdline in the discover.sh file

mikemrm (changes requested):
Looks really good! Couple of changes and we should be golden!

installer/alpine/init-aarch64
ScottGarman: Ah, good catch. I had refactored this and forgot it was still in init-aarch64.

apps/discover.sh
ScottGarman: I agree that `discover_url` is preferable to `discover_address`. And are you saying you think the default url should be `http://tinkerbell.$facility.packet.net/discover-os`? If you confirm that, I'm happy to make that change.

mikemrm: Yeah I think so, Would we be able to use the `$tinkerbell` variable instead of using the static default domain?

ScottGarman: Yep, will do!

apps/discover.sh
```diff
@@ -0,0 +1,77 @@
...
+
+arch=$(uname -m)
+packet_base_url=$(sed -nr 's|.*\bpacket_base_url=(\S+).*|\1|p' /proc/cmdline)
+facility=$(sed -nr 's|.*\bfacility=(\S+).*|\1|p' /proc/cmdline)
+tinkerbell=$(sed -nr 's|.*\btinkerbell=(\S+).*|\1|p' /proc/cmdline | sed -e 's|^http://||' -e 's|/.*||')
+phone_home_url="http://$tinkerbell/phone-home"
+
+discover_url=$(sed -nr 's|.*\bdiscover_url=(\S+).*|\1|p' /proc/cmdline)
+if [ -z "$discover_url" ]; then
+	discover_url="$tinkerbell/discover-os"
```
mikemrm: It appears $tinkerbell is just the address, so we'll need to prepend `http://` here

mikemrm (changes requested):

apps/discover.sh
ScottGarman: Fixed.

mikemrm (approved):
Looks good!

PR #151, opened 2020-10-21
correct controller id

mikemrm:
## Description

results from id-ctrl text output was base16 without the `0x` prepended,
switching to the json output returns the base10 value instead.

Signed-off-by: Mike Mason <mason@packet.com>


dustinmiller (approved):
Awesome, thank you!

dlaube (approved):

PR #156, opened 2020-10-22
Add BIOS utilities to OSIE

ScottGarman:
sum is used to update SuperMicro server BIOS and BMC firmware.


dustinmiller (approved):
Looks good to me 👍

PR #157, opened 2020-10-27
Switch f3 GRUB templates to ttyS0

dustinmiller:
## Description

Lets change the two f3 plans to utilize ttyS0 instead of ttyS1

## Why is this needed

Seems HP ships servers with this as their default TTY

## How Has This Been Tested?

Credit to @dlaube for helping sort out which TTY was in use 👍 

No break, only fix 🤞

### DIFF
--- a/grub/centos_7-f3.large.x86-grub.template
+++ b/grub/centos_7-f3.large.x86-grub.template
@@ -15,7 +15,7 @@ set default=0
 #set root='(hd0,1)'
 
 # Define console settings
-#serial console=tty0 console=ttyS1,115200n8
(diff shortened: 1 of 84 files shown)


dlaube (approved):

mikemrm (approved):

wangxin311:
f3 plan servers(HPE) COM port is configurable for VSP, you can switch between ttyS0 and ttyS1 from "Serial Port Options" configuration. So if we go with ttyS0, we need make sure when  server initializing or factory reset, the ttyS0 is been used for "Virtual Serial Port".

PR #158, opened 2020-10-29
Add ssacli and use it in smartarray_reset

dustinmiller:
## Description

Lets add in SSACLI and we'll use it for breaking apart logical drives

## Why is this needed

Logical drives get in the way of deprovisioning, so lets destroy them

## How Has This Been Tested?
Just barely

No break, only fix 🤞

### DIFF
--- a/docker/scripts/functions.sh
+++ b/docker/scripts/functions.sh
@@ -519,6 +519,23 @@ function perc_reset() {
 	fi
 }
 
+function smartarray_reset() {
+	systemmfg=$(dmidecode -s system-manufacturer | head -1)
+	echo "Adaptec hardware RAID device is present on system mfg: $systemmfg"
+
+	slots=$(ssacli ctrl all show detail | awk '/^   Slot: / {print $2}')
+
+	echo "Adaptec smart storage array, clearing logical drives"
+	for slot in $slots; do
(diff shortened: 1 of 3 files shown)


docker/scripts/functions.sh
```diff
@@ -519,6 +519,23 @@ function perc_reset() {
...
+
+	slots=$(ssacli ctrl all show detail | awk '/^   Slot: / {print $2}')
+
+	echo "HPE-SSACLI - Deleting all LDs"
+	for slot in $slots; do
+		lds=$(ssacli ctrl slot="$slot" ld all show status | awk '/^   logicaldrive / {print $2}')
+		for ld in $lds; do
+			# ssacli does not have a -yes or -force option that I can find, so yes pipe
+			# slapping `|| true` at the end to sidestep the autofail trap function
+			yes | ssacli ctrl slot="$slot" ld "$ld" delete || true
```
mikemrm: I think if we need the `|| true` part, we should have some validation afterwards to make sure all drives were removed after the fact.

mikemrm (changes requested):

docker/scripts/functions.sh
dustinmiller: Ok, looks like we found a better way for deleting the logical drives, so we won't need `|| true`

mikemrm (approved):

PR #161, opened 2020-11-11
Probe and print BIOS vendor and version on deprov

ScottGarman:
This currently only works for our Gen3 Dell and Supermicro servers as a
starting point. Eventually this will evolve into the ability to
validate BIOS configuration settings on deprov and reset them if drift
is detected.

This is WIP and I'm not sure this needs to be merged to master at all. I am however still interested in feedback on the approach itself and will test this as a custom OSIE version while doing further development.

### DIFF
--- a/docker/scripts/deprovision.sh
+++ b/docker/scripts/deprovision.sh
@@ -61,7 +61,13 @@ function autofail() {
 }
 trap autofail EXIT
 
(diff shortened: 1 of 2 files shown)


ScottGarman:
Closing, this PR is obsolete

PR #163, opened 2020-11-16
Parse instance_id from kernel cmdline rather than pulling from metadata

splaspood:
## Description

We should pass the instance id directly into osie via the kernel cmdline rather than pulling from metadata.

## Why is this needed

This is the only reason we have to query metadata from workflow-helper.sh, less overhead this way.


splaspood:
These changes were incorporated into another PR.

PR #164, opened 2020-11-19
Add deb11 and ub2104 grub templates

dustinmiller:
## Description

Lets add GRUB templates for the upcoming Debian 11 and Ubuntu 21.04 releases

## Why is this needed

Required for successful provisions

No break, only fix 🤞


dustinmiller:
#200 should resolve this

PR #167, opened 2020-11-30
Detect and report BIOS configuration drift on deprov

ScottGarman:
This is work I've been doing in another PR that I believe is ready for final review and would be low-risk to deploy to production, given that only BIOS config drift is detected and configs are not yet applied.

### DIFF
--- a/docker/scripts/functions.sh
+++ b/docker/scripts/functions.sh
@@ -98,6 +98,39 @@ function posttink() {
 	tink "POST" "${tink_host}" "${endpoint}" "${post_data}"
 }
 
+# returns a string of the BIOS vendor: "dell", "supermicro", or "unknown"
+function detect_bios_vendor() {
+	local vendor=unknown
+
+	# Check for Dell
+	if /opt/dell/srvadmin/bin/idracadm7 get BIOS.SysInformation &>/dev/null; then
+		vendor="Dell"
+	else
+		# Check for Supermicro
+		if /opt/supermicro/sum/sum -c GetDmiInfo >/dev/null; then
+			vendor="Supermicro"
+		fi
+	fi
+
+	echo "${vendor}"
+}
+
+# usage: detect_bios_version $vendor
+# returns a string of the BIOS version or "unknown" if it can't be determined.
+function detect_bios_version() {
+	local vendor=$1
+	local version=unknown
+
+	if [[ "${vendor}" == "Dell" ]]; then
+		version=$(/opt/dell/srvadmin/bin/idracadm7 get BIOS.SysInformation 2>&1 | awk -F "=" '/^#SystemBiosVersion/ {print $2}')
+	fi
+	if [[ "${vendor}" == "Supermicro" ]]; then
+		version=$(/opt/supermicro/sum/sum -c GetDmiInfo | grep --after-context 2 "^\[BIOS Information\]" | awk -F '"' '/^Version/ {print $2}')
+	fi
+
+	echo "${version}"
+}
+
 function dns_resolvers() {
 	declare -ga resolvers
 
--- a/docker/scripts/functions.sh
+++ b/docker/scripts/functions.sh
@@ -206,6 +206,41 @@ function dns_resolvers() {
 	fi
 }
 
+# usage: retrieve_current_bios_config $vendor
+# saves the current bios config to a file named current_bios.txt
+function retrieve_current_bios_config() {
+	local vendor=$1
+
+	if [[ "${vendor}" == "Dell" ]]; then
+		/opt/dell/srvadmin/bin/idracadm7 get -t json -f current_bios.txt >/dev/null
+	elif [[ "${vendor}" == "Supermicro" ]]; then
+		/opt/supermicro/sum/sum -c GetCurrentBiosCfg --file current_bios.txt >/dev/null
+	fi
+}
+
+# usage: validate_bios_config $vendor $plan
+function validate_bios_config() {
+	local vendor=$1
+	local plan=$2
+	local config_file
+
+	# Check for a BIOS config for this plan
+	config_file=$(lookup_bios_config "${plan}" "${vendor}")
+
+	if [[ -z "$config_file" ]]; then
+		echo "Unable to find a bios config file for ${plan} (${vendor}) in the manifest, skipping BIOS validation"
+		return
+	fi
+
+	if [[ ! -f "bios-configs-latest/${config_file}" ]]; then
+		echo "A config file named [${config_file}] does not exist in BIOS configs tarball, skipping BIOS validation"
+		return
+	fi
+
+	# Save current BIOS config to a local file
+	retrieve_current_bios_config "${vendor}"
+}
+
 function dns_redhat() {
 	local filename=$1
 	shift
--- a/docker/scripts/functions.sh
+++ b/docker/scripts/functions.sh
@@ -98,6 +98,23 @@ function posttink() {
 	tink "POST" "${tink_host}" "${endpoint}" "${post_data}"
 }
 
+# configures /etc/hosts so that git and git-lfs assets from GitHub and our local
+# mirror are pulled through our image cache (IPN).
+function configure_image_cache_dns() {
+	images_ip=$(getent hosts images.packet.net | awk '{print $1}')
+	cp -a /etc/hosts /etc/hosts.new
+	{
+		echo "$images_ip        github.com"
+		echo "$images_ip        github-cloud.s3.amazonaws.com"
+		echo "$images_ip        github-mirror.packet.net"
+	} >>/etc/hosts.new
+	# Note: using mv here fails (415 Unsupported Media Type) because docker sets
+	# this up as a bind mount and we can't replace it.
+	cp -f /etc/hosts.new /etc/hosts
+	echo -n "LFS pulls via github-cloud will now resolve to image cache:"
+	getent hosts github-cloud.s3.amazonaws.com | awk '{print $1}'
+}
+
 # returns a string of the BIOS vendor: "dell", "supermicro", or "unknown"
 function detect_bios_vendor() {
 	local vendor=unknown
--- a/docker/scripts/functions.sh
+++ b/docker/scripts/functions.sh
@@ -148,6 +148,54 @@ function detect_bios_version() {
 	echo "${version}"
 }
 
+# Downloads our latest BIOS configurations
+function download_bios_configs() {
+	# downloads are pulled through our image cache
+	configure_image_cache_dns
+
+	echo "Downloading latest BIOS configurations"
+	curl --fail https://github-mirror.packet.net/downloads/bios-configs-latest.tar.gz --output bios-configs-latest.tar.gz
+	curl --fail https://github-mirror.packet.net/downloads/bios-configs-latest.tar.gz.sha256 --output bios-configs-latest.tar.gz.sha256
+
+	echo "Verifying BIOS configurations tarball"
+	sha256sum --check bios-configs-latest.tar.gz.sha256
+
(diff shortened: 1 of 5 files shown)


docker/scripts/functions.sh
```diff
@@ -98,6 +98,47 @@ function posttink() {
 	tink "POST" "${tink_host}" "${endpoint}" "${post_data}"
 }
 
+# returns a string of the BIOS vendor: "dell", "supermicro", or "unknown"
+function detect_bios_vendor() {
+	set +e
+	vendor=unknown
+
+	# Check for Dell
+	dell_output=$(/opt/dell/srvadmin/bin/idracadm7 get BIOS.SysInformation 2>&1)
```
mikemrm: Could we potentially use the exit code here instead of grepping the output?

mikemrm (dismissed):

docker/scripts/functions.sh
ScottGarman: I think so. Let me investigate.

ScottGarman: Yep, this works well and is significantly cleaner. Thanks for the suggestion!

docker/scripts/functions.sh
```diff
@@ -108,6 +198,128 @@ function dns_resolvers() {
...
+		return 1
+	fi
+
+	# Delete irrelevant sections from the JSON config to normalize for diff'ing
+	jq 'del(.SystemConfiguration.ServiceTag) |
+		  del(.SystemConfiguration.TimeStamp) |
+			del(.SystemConfiguration.Components[] | select(.FQDD=="BIOS.Setup.1-1").Attributes[] | select(.Name=="SetBootOrderEn")) |
+			del(.SystemConfiguration.Components[] | select(.FQDD=="BIOS.Setup.1-1").Attributes[] | select(.Name=="BiosBootSeq")) |
+			del(.SystemConfiguration.Components[] | select(.FQDD=="iDRAC.Embedded.1").Attributes[] | select(.Name=="NIC.1#DNSRacName")) |
+			del(.SystemConfiguration.Components[] | select(.FQDD=="System.Embedded.1").Attributes[] | select(.Name=="ServerOS.1#HostName"))' \
```
mmlb: would this be easier with `gron | grep -v`?

docker/scripts/functions.sh
```diff
@@ -98,6 +98,43 @@ function posttink() {
 	tink "POST" "${tink_host}" "${endpoint}" "${post_data}"
 }
 
+# returns a string of the BIOS vendor: "dell", "supermicro", or "unknown"
+function detect_bios_vendor() {
+	set +e
```
mmlb: why does this need to be done? It looks like you handle any error paths so shouldn't need to do this.

docker/scripts/functions.sh
```diff
@@ -98,6 +98,43 @@ function posttink() {
...
+	vendor=unknown
+
+	# Check for Dell
+	if /opt/dell/srvadmin/bin/idracadm7 get BIOS.SysInformation >/dev/null 2>&1; then
```
mmlb: this can just be `&>/dev/null` which I think we use around here too

docker/scripts/functions.sh
```diff
@@ -98,6 +98,43 @@ function posttink() {
...
+function detect_bios_version() {
+	local vendor=$1
+
+	version=unknown
+	if [[ "${vendor}" == "Dell" ]]; then
+		dell_output=$(/opt/dell/srvadmin/bin/idracadm7 get BIOS.SysInformation 2>&1)
+		version=$(echo "${dell_output}" | awk -F "=" '/^#SystemBiosVersion/ {print $2}')
+	fi
+	if [[ "${vendor}" == "Supermicro" ]]; then
+		sum_output=$(/opt/supermicro/sum/sum -c GetDmiInfo)
```
mmlb: I'd do away with the `*_output` variables (which aren't being set a `local` so are polluting the global space fyi)

docker/scripts/functions.sh
```diff
@@ -98,6 +98,43 @@ function posttink() {
 	tink "POST" "${tink_host}" "${endpoint}" "${post_data}"
 }
 
+# returns a string of the BIOS vendor: "dell", "supermicro", or "unknown"
+function detect_bios_vendor() {
+	set +e
+	vendor=unknown
```
mmlb: should be `local`

docker/scripts/functions.sh
```diff
@@ -98,6 +98,43 @@ function posttink() {
...
+	set -e
+	echo "${vendor}"
+}
+
+# usage: detect_bios_version $vendor
+# returns a string of the BIOS version or "unknown" if it can't be determined.
+function detect_bios_version() {
+	local vendor=$1
+
+	version=unknown
```
mmlb: should be `local`

docker/scripts/functions.sh
```diff
@@ -135,6 +135,59 @@ function detect_bios_version() {
...
+function download_bios_configs() {
+	# downloads are pulled through our image cache
+	images_ip=$(getent hosts images.packet.net | awk '{print $1}')
+	cp -a /etc/hosts /etc/hosts.new
+	{
+		echo "$images_ip        github-cloud.s3.amazonaws.com"
+		echo "$images_ip        github.com"
+		echo "$images_ip        github-mirror.packet.net"
+	} >>/etc/hosts.new
+	cp -f /etc/hosts.new /etc/hosts
```
mmlb: can you make this into its own function so we can share with osie's fetch-image step please.

docker/scripts/functions.sh
```diff
@@ -135,6 +135,59 @@ function detect_bios_version() {
...
+
+	echo "Downloading latest BIOS configurations"
+	curl https://github-mirror.packet.net/downloads/bios-configs-latest.tar.gz --output bios-configs-latest.tar.gz
+	curl https://github-mirror.packet.net/downloads/bios-configs-latest.tar.gz.sha256 --output bios-configs-latest.tar.gz.sha256
```
mmlb: strongly encourage a `-f` here, since of we get an "http error" code curl will still return 0 and now we're running with garbage input. `-f` makes curl fail on those 4xx/5xx status.

docker/scripts/functions.sh
```diff
@@ -135,6 +135,59 @@ function detect_bios_version() {
...
+function lookup_bios_config_enforcement() {
+	local plan=$1
+	local vendor=$2
+
+	# search through the bios config manifest to extract the enforcement status for this config
+	if [[ ! -f "bios-configs-latest/manifest.txt" ]]; then
+		echo "Error: missing BIOS config manifest file"
+		return 1
+	fi
+	status=$(grep "^${plan} \+${vendor}" bios-configs-latest/manifest.txt | awk '{print $4}' || true)
```
mmlb, line 195: should have `local`

docker/scripts/functions.sh
```diff
@@ -135,6 +135,59 @@ function detect_bios_version() {
...
+function lookup_bios_config() {
+	local plan=$1
+	local vendor=$2
+
+	# search through the bios config manifest to extract the config filename
+	if [[ ! -f "bios-configs-latest/manifest.txt" ]]; then
+		echo "Error: missing BIOS config manifest file"
+		return 1
+	fi
+	configfile=$(grep "^${plan} \+${vendor}" bios-configs-latest/manifest.txt | awk '{print $3}' || true)
```
mmlb, line 179: should be `local`

docker/scripts/functions.sh
```diff
@@ -198,6 +198,40 @@ function dns_resolvers() {
...
+	fi
+}
+
+# usage: validate_bios_config $vendor $plan
+function validate_bios_config() {
+	local vendor=$1
+	local plan=$2
+
+	# Check for a BIOS config for this plan
+	config_file=$(lookup_bios_config "${plan}" "${vendor}")
```
mmlb, line 295: should be `local`

docker/scripts/functions.sh
```diff
@@ -210,6 +210,82 @@ function retrieve_current_bios_config() {
...
+	sed --in-place '/Menu name.*TOSHIBA.*order/d' "${config_file_normalized}"
+}
+
+# usage: compare_bios_config_files $vendor $config_file
+function compare_bios_config_files() {
+	local vendor=$1
+	local config_file="bios-configs-latest/$2"
+	local config_file_normalized="${config_file}.normalized"
+	local current_config="current_bios.txt"
+	local current_config_normalized="${current_config}.normalized"
```
mmlb: vendor is unused right? Also why not just pass the normalized file paths to the func?

mmlb (changes requested):

docker/scripts/functions.sh
ScottGarman: This is because we're potentially running the dell/supermicro bios utilities on platforms which don't support them, so I'm trying to prevent the script from exiting/triggering autofail due to the non-zero exit status those commands will return.

mmlb: Thats ok though, you won't have bash exit because you handle all the error cases:
```
[manny@dellnix:~]$ bash <<EOF
set -e
if /opt/dell/srvadmin/bin/idracadm7; then                           
  echo we should not see this, there is no idracadm7
elif /opt/supermicro/sum/sum; then
  echo we should not see this, there is no sum tool
else
  echo ok so everything errored, its ok bash does not die because we handled it with if and else
fi
EOF
bash: line 2: /opt/dell/srvadmin/bin/idracadm7: No such file or directory
bash: line 4: /opt/supermicro/sum/sum: No such file or directory
ok so everything errored, its ok bash does not die because we handled it with if and else
```
This shouldn't trigger the exiting/autofail either.

ScottGarman: Ah, it is now since I'm no longer using intermediate variables to store/parse the output of the commands, which was my previous approach. Good catch.

docker/scripts/functions.sh
ScottGarman: Great idea, I'm looking into adding this tonight.

docker/scripts/functions.sh
ScottGarman: Fixed.

docker/scripts/functions.sh
ScottGarman: Fixed.

docker/scripts/functions.sh
ScottGarman: Fixed.

docker/scripts/functions.sh
ScottGarman: Added (in long form of course, --fail).

docker/scripts/functions.sh
ScottGarman: Fixed.

docker/scripts/functions.sh
ScottGarman: Fixed.

docker/scripts/functions.sh
ScottGarman: Fixed.

docker/scripts/functions.sh
ScottGarman: Yep, vendor was an artifact from before I broke out the vendor-specific normalization code into its own functions. And I've specified the bios-configs-latest/ file path in the function call rather than inside the function.

docker/scripts/functions.sh
mmlb: of course! :+1:

docker/scripts/functions.sh
ScottGarman: I've added the gron binary to OSIE, will hold off on refactoring to use it just yet as we discussed earlier.

mmlb (approved):

PR #173, opened 2020-12-12
Gracefully handle failures during BIOS config dump

ScottGarman:
There can be circumstances were the command to dump the BIOS config
might fail. Handle these with a warning log and not with a deprov
failure.

This PR also improves some of the autofail message resolution during BIOS validation.


jgavinray (approved):
I approve of this message.

PR #175, opened 2020-12-16
Fix caddyserver repo URL in CI

ScottGarman:
Looks like mholt's repository has gone private - we should be using the
official caddy repo instead anyway.


mmlb (dismissed):

mmlb (approved):
:tada: lgtm

mmlb:
closing in favor of #177 since I can't push to your branch Scott :/

PR #177, opened 2020-12-18
ci: build a custom proxy binary using examples from caddyserver and h…

mmlb:
## Description

mholt/caddy is no longer redirecting to caddyserver/caddy and thus our CI build is breaking

## Why is this needed

CI uses caddy to proxy fetches from EM production urls and makes use of the caddy.upload (since renamed to http.upload) plugin to handle the POST/PUT requests from OSIE. Since mholt/caddy went private we need to update our build steps. This also updates to caddy v1 and http.upload v3 so hopefully things stay stable for a while more.

## How Has This Been Tested?

Caddy runs locally + in CI

CI is un-broken, PRs can be merged, joy may be felt once again.


CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 # Changelog
+
```
ScottGarman: I assume you were just triggering CI with this change and want to remove this commit from the PR

mmlb: yep that came in from some testing I had in master and was confusing me quite a bit :D

ScottGarman (approved):

PR #178, opened 2020-12-23
Skip BIOS validation process for BIOS configs tarball errors

ScottGarman:
Since we download the BIOS configs tarball via IPN and it's a small file
to begin with (under 1 MB), download failures could indicate a
networking issue. It is recommended to configure log alerts based on
these error messages.

### DIFF
--- a/docker/scripts/functions.sh
+++ b/docker/scripts/functions.sh
@@ -154,14 +154,26 @@ function download_bios_configs() {
 	configure_image_cache_dns
 
 	echo "Downloading latest BIOS configurations"
-	curl --fail https://github-mirror.packet.net/downloads/bios-configs-latest.tar.gz --output bios-configs-latest.tar.gz
-	curl --fail https://github-mirror.packet.net/downloads/bios-configs-latest.tar.gz.sha256 --output bios-configs-latest.tar.gz.sha256
+	if ! curl --fail https://github-mirror.packet.net/downloads/bios-configs-latest.tar.gz --output bios-configs-latest.tar.gz; then
+		echo "Error: Download of BIOS configs tarball failed"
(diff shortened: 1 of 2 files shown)


docker/scripts/functions.sh
```diff
@@ -154,14 +154,26 @@ function download_bios_configs() {
...
-	sha256sum --check bios-configs-latest.tar.gz.sha256
+	if ! sha256sum --check bios-configs-latest.tar.gz.sha256; then
+		echo "Error: Verifying checksum of BIOS configs tarball failed"
+		return 1
+	fi
 
 	echo "Extracting BIOS configurations tarball"
-	tar -zxf bios-configs-latest.tar.gz
+	if ! tar -zxf bios-configs-latest.tar.gz; then
+		echo "Error: Extracting BIOS configs tarball failed"
```
mmlb, line 174: I'm not sure all these `if` and echos in the error case are really necessary. I feel like we already have enough info with the pre message that if there's an error that is caught by `set -e` we'd be able to figure things out. Otherwise why don't we do `if ! ...` for every line of code too?

ScottGarman: The point of this is to catch errors so we can report them as warnings and then skip the BIOS validation behavior, without triggering the autofail and failing the deprovision.

mmlb: sigh bash craziness bites me again... I now am aware that `set -e` is totally and globally ignored when part of complex pipelines even in sub function calls. So -e isn't actually on here since this is called in an if...

mmlb (approved):

ScottGarman:
Closing, this has been superceded by PR #179

PR #179, opened 2021-01-07
Skip BIOS validation for reserved hardware and check manifest for enforcement status

ScottGarman:
Reserved hardware can have custom BIOS configs that we don't want to
touch. At some point we'll want to have finer grained metadata on what
BIOS config to use as a source of truth, but for now we simply won't
apply BIOS configs to reserved hardware devices.

In the bios configs tarball we have a manifest.txt file which includes
an enforcement status field. If this field is not set to "enforce", all
BIOS config checks will happen, but the config will not be applied. This
allows us to improve the config normalization filters during a
development period, after which we can set the enforcement status to
enable the config to be written when drift is detected.

### DIFF
--- a/docker/scripts/deprovision.sh
+++ b/docker/scripts/deprovision.sh
@@ -37,6 +37,7 @@ assert_all_args_consumed "$OPTIND" "$@"
 
 declare facility && set_from_metadata facility 'facility' <"$metadata"
 declare class && set_from_metadata class 'class' <"$metadata"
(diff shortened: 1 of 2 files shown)


markjacksonfishing (dismissed):

ScottGarman:
Note: I've tested the reserved status by modifying kant's data directly, and so I've verified the behavior works when reserved is true vs. false.

markjacksonfishing (approved):

PR #188, opened 2021-01-29
Do not fail deprovisions when SOL causes docker to hang

ScottGarman:
If SOL isn't working properly during deprovision, docker can hang even
though the deprovision script has run to completion. This change sets
the autofail message to "deprovision completed" at the very end of the
deprovision script, so that if docker hangs and is killed by the timeout
watcher, we can detect that the deprovision script completed and cleanup
gracefully instead of failing the deprovision.

Signed-off-by: Scott Garman <sgarman@packet.com>

### DIFF
--- a/apps/osie-installer.sh
+++ b/apps/osie-installer.sh
@@ -2,6 +2,15 @@
 
 # shellcheck shell=dash
 
+cleanup() {
+	# shellcheck disable=SC2169
+	if [ -x "$statedir/cleanup.sh" ]; then
+		reason='running cleanup.sh'
+		cd "$statedir"
+		exec ./cleanup.sh
+	fi
+}
+
 reason='unknown'
 fail() {
 	if [ "$reason" = "docker exited with an error (osie-installer)" ]; then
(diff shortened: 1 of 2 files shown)


markjacksonfishing (approved):

apps/osie-installer.sh
```diff
@@ -11,6 +20,14 @@ fail() {
 		# before the timeout occurred.
 		if [ -f "${statedir}/autofail_stage" ]; then
 			stage=$(cat "${statedir}/autofail_stage")
+			# If SOL isn't working properly on deprovision, docker can hang even though
+			# the deprovision script has run to completion. Detect this by checking the
+			# autofail_stage value, which would indicate the deprovison completed. Don't
+			# report failure in that case, but just run the cleanup() function and exit.
```
mmlb, line 26: docker doesn't actually hang right? docker exits with an error and so the script exits and then this trap is reached and we phone back with a timed out error. I think thats the case at least, if so can you update the comment to match.

apps/osie-installer.sh
```diff
@@ -11,6 +20,14 @@ fail() {
...
+			if [ "$stage" = "deprovision completed" ]; then
```
mmlb, line 27: we shouldn't tie this to just deprovision, how about just handling `stage=completed`? And updating the other scripts to finish up with `completed`

mmlb (changes requested):

dlaube:
> If SOL isn't working properly during deprovision, docker can hang even
> though the deprovision script has run to completion. This change sets
> the autofail message to "deprovision completed" at the very end of the
> deprovision script, so that if docker hangs and is killed by the timeout
> watcher, we can detect that the deprovision script completed and cleanup
> gracefully instead of failing the deprovision.
> 
> Signed-off-by: Scott Garman [sgarman@packet.com](mailto:sgarman@packet.com)

We are seeing this problem on OSIE provisions as well. I have tested this build on a new c3.medium open19 machine in NY5 and the same failure mode is observed. The error received even with this build is "Timed out during OSIE final stage" All other steps appear successful and if we reboot the machine manually, it makes it successfully into the target OS fine.

mmlb:
I've tweaked this to also work for more than just deprovisions and opened up #198 because I can't push to your branch @ScottGarman :/

PR #196, opened 2021-03-02
Add blocksize for specific Intel NVMe drives

dlaube:
## Description

This fixes the namespare error while deprovisioning machines with specific Intel NVMe drives

Fixes: #

## How Has This Been Tested?
Testing via custom userdata / custom OSIE method 

NA

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/docker/scripts/deprovision.sh
+++ b/docker/scripts/deprovision.sh
@@ -163,7 +163,7 @@ if [[ $preserve_data == false ]]; then
 				done
 			fi
 
-			nvmemodel=$(nvme id-ctrl "$drive" -o json | jq -r '.mn')
+			nvmemodel=$(nvme id-ctrl "$drive" -o json | jq -r '.mn' | sed -e 's/[[:space:]]*$//')
 			if [[ $nvmemodel == 'INTEL SSDPE2KX040T8' ]]; then
 				# Set specific block size depending on physical BD
 				sectors=$((max_bytes / 4096))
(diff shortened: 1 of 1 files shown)


dlaube:
Tested against existing server types with NVMe for regression testing and all is good.

docker/scripts/deprovision.sh
```diff
@@ -163,8 +163,15 @@ if [[ $preserve_data == false ]]; then
...
 
-			# flbas 0 uses 512 byte sector sizes
-			sectors=$((max_bytes / 512))
+			nvmemodel=$(nvme id-ctrl "$drive" -o json | jq -r '.mn' | sed -e 's/[[:space:]]*$//')
+			if [[ $nvmemodel == 'INTEL SSDPE2KX040T8' ]]; then
+				# Set specific block size depending on physical BD
+				sectors=$((max_bytes / 4096))
+			else
+				# default flbas 0 uses 512 byte sector sizes
+				sectors=$((max_bytes / 512))
```
mmlb, line 172: nvme tool doesn't have a block size output we can use instead of hard coding cards?

dlaube: My thoughts exactly. The nvme tool does have it but its something that gets removed at deprov. I reluctant to hard code the disk model (like we still do for HW raid controllers) but this is what we have right now that works.

mmlb: how do we get it back? Do we need to create a namespace first?

dustinmiller (approved):
Works for now until we can find a better method :)

PR #198, opened 2021-03-08
Do not fail when SOL causes docker to return error

mmlb:
## Description

Don't blindly fail just because docker returned an error.

## Why is this needed

Defensive coding to protect against effectively fake failures for production EM use, mostly deprovisions but also some provisions.

## How Has This Been Tested?

In production.

### DIFF
--- a/apps/osie-installer.sh
+++ b/apps/osie-installer.sh
@@ -2,6 +2,15 @@
 
 # shellcheck shell=dash
 
+cleanup() {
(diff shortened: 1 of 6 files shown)


mmlb:
Using this to replace #188 as I can't push to that branch :/

ScottGarman (dismissed):

ScottGarman (approved):

PR #201, opened 2021-03-18
osie: try regular git now

dustinmiller:
## Description

We've been building a custom git for a long time, but seems to me like we no longer need to be.

## Why is this needed

The build process for the aforementioned git is deteriorating/broken at the moment

## How Has This Been Tested?

I tested this on many of our plan types via custom OSIE deployments


gemagomez (approved):
LGTM

dlaube (approved):

PR #203, opened 2021-03-20
osie: upgrade to ubuntu:focal

dustinmiller:
## Description

Lets upgrade OSIE to Ubuntu 20.04

## Why is this needed

This PR upgrades the OSIE container from Xenial to Focal- this lets us
build less packages and use more of the included software instead. This
also moves away from several of the slower build processes. Build times
are down from ~39/12minutes (arm/x86) to ~9/7minutes as a result

## How Has This Been Tested?
Provs/deprovs on the majority of x86 at this time, more testing in progress


dustinmiller:
only issue I see on ARM UEFI installs
```+ mount -o remount,ro /sys/firmware/efi/efivars                                                                                                                                                                    
+ grub-install --recheck --bootloader-id=GRUB --root-directory=/mnt/target --efi-directory=/mnt/target/boot/efi                                                                                                    
Installing for arm64-efi platform.                                                                       
grub-install: warning: Cannot set EFI variable Boot0000.                                                 
grub-install: warning: efivarfs_set_variable: failed to create /sys/firmware/efi/efivars/Boot0000-8be4df61-93ca-11d2-aa0d-00e098032b8c for writing: Read-only file system.
grub-install: warning: _efi_set_variable_mode: ops->set_variable() failed: Read-only file system.        
grub-install: error: failed to register the EFI boot entry: Read-only file system.                       
+ autofail      ```

mmlb:
hmm, works on x86_64 uefi?

tstromberg:
It looks like this draft PR is 3 months old - is it ready to be reviewed yet or should it be closed?

PR #204, opened 2021-03-23
Always pull the latest tink-worker image

mfranczy:
## Description

Change tink-worker image to the latest tag.

## Why is this needed

Currently if somebody will run the latest sandbox then they will not be able to pull tink-worker image.
As the comments indicates: [link](https://github.com/tinkerbell/sandbox/blob/master/setup.sh#L390)

> osie looks for tink-worker:latest, so we have to play with it a bit

The above sentence is false righ now, OSIE points to the specific sha.

Fixes: #

## How Has This Been Tested?
Local setup.

Won't be able to download tink-worker image inside OSIE.

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/apps/workflow-helper.sh
+++ b/apps/workflow-helper.sh
@@ -12,7 +12,7 @@ syslog_host=$(sed -nr 's|.*\bsyslog_host=(\S+).*|\1|p' /proc/cmdline)
 worker_id=$(sed -nr 's|.*\bworker_id=(\S+).*|\1|p' /proc/cmdline)
 instance_id=$(sed -nr 's|.*\binstance_id=(\S+).*|\1|p' /proc/cmdline)
 
-tink_worker_image="${docker_registry}/tinkerbell/tink-worker:sha-5e1f0fd8"
+tink_worker_image="${docker_registry}/tinkerbell/tink-worker:latest"
 
 # Create workflow motd
 cat <<'EOF'


markjacksonfishing (commented):
I am of the mindset that we should avoid using the :latest tag when deploying containers, because this makes it hard to track which version of the image is running and hard to roll back

mfranczy:
> I am of the mindset that we should avoid using the :latest tag when deploying containers, because this makes it hard to track which version of the image is running and hard to roll back

@markyjackson-taulia I totally agree with that but you track versions (at least in sandbox) here https://github.com/tinkerbell/sandbox/blob/master/current_versions.sh later the tink-worker image is retagged to local repo and latest tag. Without this change it's not possible to download tink-worker to OSIE.

Look at the comment here https://github.com/tinkerbell/sandbox/blob/master/setup.sh#L390, this PR https://github.com/tinkerbell/osie/pull/192/files#diff-3b3d4f0020350fb13779da03f9573a5b3f4edb54dc62ece89db157d50972440bL81 broke the assumption that was before.

I agree this change makes things implict but at least it fixes the OS provisioning.
So either it must be fixed here or synced somehow with the [current_versions](https://github.com/tinkerbell/sandbox/blob/master/current_versions.sh)

btw this pr https://github.com/tinkerbell/osie/pull/192/files breaks other things too, but maybe I will elaborate on that in a different issue or slack channel (I have to confirm first a few things).

PR #212, opened 2021-04-27
Route new GitHub LFS hostname via IPN, increase github-mirror health check timeout

ScottGarman:
GitHub appears to have changed their LFS backend to use a new hostname,
github-cloud.githubusercontent.com. In order to route this via IPN for
caching purposes, add it to the /etc/hosts file using the anycast IP for
IPN.
    
It's not yet confirmed whether they've stopped using the AWS s3 hostname,
so we're leaving that in place for the time being.

Also:

Increase the github-mirror connectivty check timeout
    
60 seconds should be long enough to handle transient slowdowns of
github-mirror that may occur during repo sync operations with our
packet-images GitHub LFS repo.

### DIFF
--- a/docker/scripts/functions.sh
+++ b/docker/scripts/functions.sh
@@ -106,6 +106,7 @@ function configure_image_cache_dns() {
 	cp -a /etc/hosts /etc/hosts.new
 	{
 		echo "$images_ip        github.com"
+		echo "$images_ip        github-cloud.githubusercontent.com"
 		echo "$images_ip        github-cloud.s3.amazonaws.com"
 		echo "$images_ip        github-mirror.packet.net"
 	} >>/etc/hosts.new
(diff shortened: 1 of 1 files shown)


ScottGarman:
I've tested this extensively as a custom OSIE, and it appears to be working as intended.

mmlb (approved):

markjacksonfishing (approved):

splaspood (approved):

docker/scripts/functions.sh
```diff
@@ -106,14 +106,15 @@ function configure_image_cache_dns() {
...
 		echo "$images_ip        github.com"
+		echo "$images_ip        github-cloud.githubusercontent.com"
 		echo "$images_ip        github-cloud.s3.amazonaws.com"
 		echo "$images_ip        github-mirror.packet.net"
 	} >>/etc/hosts.new
 	# Note: using mv here fails (415 Unsupported Media Type) because docker sets
 	# this up as a bind mount and we can't replace it.
 	cp -f /etc/hosts.new /etc/hosts
 	echo -n "LFS pulls via github-cloud will now resolve to image cache:"
-	getent hosts github-cloud.s3.amazonaws.com | awk '{print $1}'
```
tstromberg, line 116: I may be misreading the PR description, but I feel like this line removal feels counter to it:

`It's not yet confirmed whether they've stopped using the AWS s3 hostname, so we're leaving that in place for the time being.`

ScottGarman: This particular line is just printing what the IP is that github-cloud.s3.amazonaws.com/github-cloud.githubusercontent.com is currently pointing to. Just above this, we hard-wire this to use `$images_ip` in /etc/hosts, so all this is doing is printing a verification that this hostname is resolving to our anycast IP for IPN.

PR #219, opened 2021-05-20
Change bootloader-id to work correctly after grub boothole v2 updates

mmlb:
## Description

Makes osie install uefi with --bootloader-id=ubuntu.

## Why is this needed

Recent Ubuntu grub updates to fix the BootHole vulnerabilities ended up
hard coding the grub config lookup from \efi\ubuntu, but `grub-install --bootloader-id=grub`
places it into `\efi\grub`. Boot failures ensue.

## How Has This Been Tested?

Modified `ci/vm.sh` to boot the installed VM.
I was able to reproduce the boot hang without this change and no issues after.

### DIFF
--- a/docker/scripts/grub-installer.sh
+++ b/docker/scripts/grub-installer.sh
@@ -2,10 +2,11 @@
 
 # shellcheck disable=SC1091
 source functions.sh && init
+set -u
 
 USAGE="Usage: $0 -t /mnt/target -C /path/to/cprout.json
 Required Arguments:
-	-p plan      Server plan (ex: t1.small.x86)
+	-p class     Server class (ex: t1.small.x86)
 	-t target    Target mount point to write configs to
 	-C path      Path to file containing cpr.sh output json
 	-D path      Path to grub.default template
@@ -19,7 +20,7 @@ Description: This script will configure grub for the target distro
 "
 while getopts "p:t:C:D:T:hv" OPTION; do
 	case $OPTION in
-	p) plan=$OPTARG ;;
+	p) class=$OPTARG ;;
 	t) target="$OPTARG" ;;
 	C) cprout=$OPTARG ;;
 	D) default_path=$OPTARG ;;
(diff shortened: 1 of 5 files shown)


mmlb:
Failing for centos_7 atm

docker/scripts/grub-installer.sh
```diff
@@ -141,7 +141,7 @@ install_grub_osie() {
 			install -Dm755 "$grubefi" "$target/boot/efi/EFI/GRUB2/GRUBAA64.EFI"
 		else
 			# grub-install doesn't error if efibootmgr can't actually set the boot entries/order
-			efibootmgr | tee /dev/stderr | grep -iq grub
+			efibootmgr | tee /dev/stderr | grep -iq ubuntu
```
tstromberg: Neat! I've never seen the `tee /dev/stderr` trick. The comment I had was this can be fragile, and break in certain obscure environments (chroots and what not). The native way to do this is in shells (at least those produced after 1993) is:

`efibootmgr 2>&1 | grep -iq ubuntu`

Feel free to ignore this suggestion in preference to putting it into a refactor-only PR.

docker/scripts/grub-installer.sh
```diff
@@ -126,7 +126,7 @@ install_grub_osie() {
 	else
 		[[ $arch == aarch64 ]] && mount -o remount,ro /sys/firmware/efi/efivars
 
-		grub-install --recheck --bootloader-id=GRUB --root-directory="$target" --efi-directory="$target/boot/efi"
+		grub-install --recheck --bootloader-id=ubuntu --root-directory="$target" --efi-directory="$target/boot/efi"
```
tstromberg: Since it seems like the bootloader_id has to exactly match between here and the later `efibootmgr` line later - could you please extract it into a variable?

`readonly BOOTLOADER_ID=ubuntu`

tstromberg (changes requested):

docker/scripts/grub-installer.sh
mmlb: The `tee` is here so we can see all of `efibootmgr`'s output on stderr while copying to stdout and into `grep`. `efitbootmgr 2>&1` won't duplicate to stderr. This is really just a hack to get around running `efibootmgr` twice, which doesn't really buy us anything really, I'm fine with calling it twice.

docker/scripts/grub-installer.sh
mmlb: Not sure if this is going to stay this way to get centos (and presumably debian and others) working. I'll get rid of the duplication if it happens to stick around though.

tstromberg:
Is this PR ready to be re-reviewed?

mmlb:
Nope, closing in favor of #228

PR #221, opened 2021-06-02
Shrink saved docker images

### LINKED ISSUE #2: reduce osie size
We should remove unnecessary packages, perhaps package x86_64/aarch64 architectures independently, and generally work to slim down OSIE substantially.

mmlb:
## Description

Shrink the saved/exported docker size by squashing and compressing.

## Why is this needed

Images are larger than necessary, lets save some bits.

Fixes #2

## How Has This Been Tested?

VM tests

Smaller fetches from the internet, quicker docker load times.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -3,14 +3,6 @@ FROM ubuntu:xenial
 VOLUME /statedir
 ENTRYPOINT ["/entrypoint.sh"]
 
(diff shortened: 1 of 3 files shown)


markjacksonfishing (approved):
/lgtm

splaspood (approved):

PR #222, opened 2021-06-02
Test deprovision via VM

mmlb:
## Description

This PR adds deprovision to the vm based test, along with preparations for
another PR I will open that adds booting the VM and checking for phone-home.

## Why is this needed

Deprovision isn't currently tested.

## How Has This Been Tested?

Qemu based tests in ci/vm.sh

### DIFF
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -21,6 +21,7 @@ RUN apk add --no-cache --no-progress --update --upgrade \
         libarchive-tools \
         make \
         musl-dev \
+        ncurses \
         ovmf \
         pigz \
         py3-pip \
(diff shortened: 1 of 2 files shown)


splaspood (dismissed):

markjacksonfishing (dismissed):
Do we squash? I think this should be

mmlb:
> Do we squash? I think this should be

Never!

mmlb:
Rebased and forced push to fix a typo (virtion-net) in a commit message and fixed the shellcheck warnings.

markjacksonfishing (dismissed):

mmlb:
I've got a boot test that I'm trying to add to this PR, its expected to fail for now (follow up PR will fix) but its just hanging. So I'm debugging that.

mmlb:
> I've got a boot test that I'm trying to add to this PR, its expected to fail for now (follow up PR will fix) but its just hanging. So I'm debugging that.

NM, I'll do that in a different PR.

markjacksonfishing (approved):
hey, I just reviewed this.
And this is crazy,
but here's my hogwash,
so squash this maybe

PR #223, opened 2021-06-03
Multistage Docker Builds

### LINKED ISSUE #129: OSIE image should use multi stage builds
We build a few packages from source, some of them may go away with #128 but some are likely to still be needed. Using multi stage builds would allow making better use layer caching and likely help with the final image size (#2).

mmlb:
## Description

Changes the osie container to use multistage builds for every binary built from source.

## Why is this needed

Make builds faster. Fixes #129

## How Has This Been Tested?

Builds run, all the tools are there and can be invoked.


docker/Dockerfile
```diff
@@ -1,4 +1,34 @@
-FROM ubuntu:xenial
+FROM ubuntu:xenial as builder
```
markjacksonfishing, line 1: Should we be pinning a version here?

docker/Dockerfile
```diff
@@ -1,4 +1,34 @@
...
+RUN apt-get update -y && apt-get install -y build-essential curl
```
markjacksonfishing, line 2: `--no-install-recommends` should also be in here

docker/Dockerfile
```diff
@@ -1,4 +1,34 @@
...
+
+FROM builder as racadm
+RUN apt-get update && apt-get install -y alien wget
+RUN rpm --verbose --import https://linux.dell.com/repo/pgp_pubkeys/0x1285491434D8786F.asc
+WORKDIR /tmp/osie
+RUN wget \
+        https://dl.dell.com/FOLDER05920767M/1/DellEMC-iDRACTools-Web-LX-9.4.0-3732_A00.tar.gz \
+        http://linux.dell.com/repo/community/openmanage/940/bionic/pool/main/s/srvadmin-omilcore/srvadmin-omilcore_9.4.0_amd64.deb
+RUN tar -xvf DellEMC-iDRACTools-Web-LX-9.4.0-3732_A00.tar.gz
+RUN alien iDRACTools/racadm/RHEL8/x86_64/*.rpm
```
markjacksonfishing: Use ./*glob* or -- *glob* so names with dashes won't become options.

markjacksonfishing (changes requested):
I know this is still in draft but some initial comments

docker/Dockerfile
mmlb: that doesn't apply hear since the `iDrac...` does the same effect as `./`

docker/Dockerfile
mmlb: I'll give this a try.

docker/Dockerfile
mmlb: I don't think its necessary. Doesn't buy us much. The apt caches are still blown away so we'd still need to do an `apt-get update` for most of the things we want. I've got a jira ticket to possibly explore pinning the version to a date and what not but that'll be left for another day.

docker/Dockerfile
mmlb: Nope that doesn't work, and seeing as this is just the build image I'm ok with having some extra junk in there.

markjacksonfishing (dismissed):

mmlb:
This is in the back burner for the moment, need to figure out how to squash the `buildx` built images.

nshalman:
@mmlb should we keep this open?

jacobweinstock:
Discussed this with @mmlb in the Triage party and he will re-open this if/when desired/needed.

PR #224, opened 2021-06-03
Run packet-hardware inventorybios during deprov

ScottGarman:
On supported platforms (currently Dell servers), this will output the
state of a handful of BIOS features and their state. For now we'll just
echo the JSON output to get it into the logs. Later on, the data will be
posted to a forthcoming data store.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -68,7 +68,7 @@ RUN apt-get update -y && \
     apt-get -qy clean && \
     rm -rf /var/lib/apt/lists/* /tmp/osie/
 
-ARG PACKET_HARDWARE_COMMIT=68a25da2b6a3d3281838bff6b62639438116d0d1
+ARG PACKET_HARDWARE_COMMIT=ddbafcbc74ef3db0677d56733442cd9f6f76a317
 ARG PACKET_NETWORKING_COMMIT=06061801efaf34caa8d8b1ae5973c15e5a23659d
 
 RUN curl https://bootstrap.pypa.io/pip/3.5/get-pip.py | python3 && \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -8,7 +8,7 @@ ENV LANG=C.UTF-8
 
 # apt sources fix
 COPY sources.list.aarch64 /tmp/osie/
-RUN if [ $(uname -m) = 'aarch64' ]; then \
+RUN if [ "$(uname -m)" = 'aarch64' ]; then \
         mv /tmp/osie/sources.list.aarch64 /etc/apt/sources.list; \
     fi && \
     rm -rf /tmp/osie
@@ -23,7 +23,7 @@ RUN apt-get update -y && \
 
 # gron - grep for JSON
 COPY lfs/gron-0.6.0-amd64 /tmp/osie/
-RUN if [ $(uname -m) = 'x86_64' ]; then \
+RUN if [ "$(uname -m)" = 'x86_64' ]; then \
         mv /tmp/osie/gron-0.6.0-amd64 /usr/bin/gron; \
     fi && \
     rm -rf /tmp/osie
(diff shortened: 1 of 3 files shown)


mmlb (dismissed):

markjacksonfishing (dismissed):

markjacksonfishing (dismissed):

docker/Dockerfile
```diff
@@ -68,7 +68,15 @@ RUN apt-get update -y && \
     apt-get -qy clean && \
     rm -rf /var/lib/apt/lists/* /tmp/osie/
 
-ARG PACKET_HARDWARE_COMMIT=68a25da2b6a3d3281838bff6b62639438116d0d1
+# ironlib cli wrapper, a prereq for packet-hardware since we don't run it from its container
+COPY lfs/getbiosconfig /tmp/osie/
+RUN cd /tmp/osie && \
+    if [ $(uname -m) = 'x86_64' ]; then \
```
tstromberg: Always quote strings containing variables, command substitutions, spaces or shell meta characters, unless careful unquoted expansion is required or it’s a shell-internal integer 

Stolen from https://google.github.io/styleguide/shellguide.html - but shellcheck also finds this: https://github.com/koalaman/shellcheck/wiki/SC2046

docker/scripts/deprovision.sh
```diff
@@ -64,7 +64,7 @@ function autofail() {
 trap autofail EXIT
 
 # Check BIOS config and update if drift is detected
-if [[ $arch == x86_64 ]] && [[ $reserved != "true" ]]; then
+if [[ $arch == "x86_64" ]] && [[ $reserved != "true" ]]; then
```
tstromberg, line 67: Quote your variables here too to avoid expansion hell.

docker/scripts/deprovision.sh
```diff
@@ -199,7 +199,7 @@ if [[ $preserve_data == false ]]; then
 	# LSI MegaRAID and Dell PERC series 9
 	set_autofail_stage "checking/resetting MegaRAID/PERC RAID controllers"
 	# do not do grep -q, it doesn't play well with pipefail when lots of pci devices exist
-	if lspci -nn | grep -v 'SAS3008' | grep LSI >/dev/null && [[ $arch == x86_64 ]]; then
+	if lspci -nn | grep -v 'SAS3008' | grep LSI >/dev/null && [[ $arch == "x86_64" ]]; then
```
tstromberg: Do the cheapest check first -- it isn't necessary to run lspci unless arch is x86_64.

docker/scripts/deprovision.sh
```diff
@@ -213,7 +213,7 @@ if [[ $preserve_data == false ]]; then
 
 	# Marvell (Dell) BOSS-S1
 	set_autofail_stage "checking/resetting Marvell RAID controllers"
-	if lspci -nn | grep '88SE9230' | grep Marvell >/dev/null && [[ $arch == x86_64 ]]; then
+	if lspci -nn | grep '88SE9230' | grep Marvell >/dev/null && [[ $arch == "x86_64" ]]; then
```
tstromberg: Check platform first here as well.

docker/scripts/deprovision.sh
```diff
@@ -271,12 +271,21 @@ baremetal_2a2 | baremetal_2a4 | baremetal_hua)
 *)
 	set_autofail_stage "running packet-hardware inventory"
 	packet-hardware inventory --verbose --tinkerbell "${tinkerbell}/hardware-components"
+	if [[ $arch == "x86_64" ]]; then
```
tstromberg, line 275: Can you add a comment as to why the bios is only inventoried for x86-64? It isn't obvious from the context why, so this behavior might get refactored out later.

tstromberg (changes requested):

docker/Dockerfile
ScottGarman: Agreed. You're catching things that are an issue throughout the Dockerfile, so I've updated them everywhere for consistency.

docker/scripts/deprovision.sh
ScottGarman: Done (also fixed throughout the file for consistency). For whatever reason, shellcheck isn't complaining about this, so it's been missed.

docker/scripts/deprovision.sh
ScottGarman: Nice catch - fixed.

docker/scripts/deprovision.sh
ScottGarman: Fixed as well.

docker/scripts/deprovision.sh
ScottGarman: Done.

markjacksonfishing (approved):

Issue #226, opened 2021-06-07
osie.sh based installs fail to boot


mmlb:
Installations done through osie.sh (legacy stuff) fail to boot since it uses Ubuntu's grub not the grub embedded in the image. Ubuntu's grub has been updated to mitigate against [BootHole round 2](https://ubuntu.com/blog/mitigating-boothole-theres-a-hole-in-the-boot-cve-2020-10713-and-related-vulnerabilities) issues. Ubuntu's grub has a hard coded path to the grub.cfg which is not where the default grub-install puts it and thus we fail to boot on powerup.

## Expected Behaviour

Legacy installations have grub properly installed, able to find the configuration file and thus boot from disk.

## Current Behaviour

Grub can't find it's configuration file,and hangs the boot.

## Possible Solution

Going to re-work grub installation to install via chroot, using the image's grub.

jacobweinstock:
Archiving repo, closing issue per [GitHub's recommendation](https://docs.github.com/en/repositories/archiving-a-github-repository/archiving-repositories#about-repository-archival).

PR #227, opened 2021-06-07
Test VM poweron, boot, and phonehome

mmlb:
## Description

Add power on + boot from disk + expect phone home test to the VM based tests.

## Why is this needed

We hit #226 in production because we don't boot test osie.sh installations.

## How Has This Been Tested?

`make test-x86_64 test_aarch64`.

Legacy code, only EM is known to be impacted.

## Checklist:

I have:

- [x] added unit or e2e tests

### DIFF
--- a/ci/vm.sh
+++ b/ci/vm.sh
@@ -357,6 +357,13 @@ do_test() {
 	rm -f uploads/*
 
 	color=34
+	colorize $color "== Running Boot & Phone-Home Test =="
+	test_boot_and_phone_home |& stdbuf -i 0 sed "s/^/$(colorize $color 'test_boot_and_phone_home│')/" &&
+		echo "this test is expected to fail" >&2 &&
+		exit 1
+	rm -f uploads/*
+
+	color=35
 	colorize $color "== Running Deprovision Test =="
 	test_deprovision |& stdbuf -i 0 sed "s/^/$(colorize $color 'test_deprovision│')/"
 	rm -f uploads/*
@@ -406,6 +413,44 @@ test_provision() {
 	got=$(grep -hr "$eventid" uploads)
 	#shellcheck disable=SC2089
 	local want='{"type":"provisioning.109"}'
+	diff -u <(jq -cS . <<<"$want") <(jq -cS . <<<"$got")
+}
+
+test_boot_and_phone_home() {
+	run_vm &
+	local vmpid=$!
+
+	local i=0
+	until grep -qr instance_id uploads; do
+		# timeout after 2min
+		if ((i++ == 12)); then
+			break
+		fi
+		sleep 10
+	done
+	echo 'system_powerdown' | socat unix-connect:monitor.sock -
+
+	i=0
+	local qpid
+	# wait 2min for the vm to shutdown
+	until ((i++ == 12)); do
+		qpid=$(pgrep qemu || :)
+		if [[ -z ${qpid:-} ]] || ((qpid != vmpid)); then
+			break
+		fi
+		sleep 10
+	done
+	if ((i >= 12)); then
+		kill -KILL $vmpid
+		sleep 1
+	fi
+
+	# check for phone-home
+	local eventid=instance_id
+	grep -qr "$eventid" uploads
+	local got
+	got=$(grep -hr "$eventid" uploads)
+	local want='{"instance_id":"'"$id"'"}'
 	diff -u <(jq -cS . <<<"$want") <(jq -cS . <<<"$got")
 }
 
--- a/ci/vm.sh
+++ b/ci/vm.sh
@@ -113,7 +113,7 @@ gen_metadata() {
 		      }
 		    ],
 		    "bonding": {
-		      "mode": 5
+		      "link_aggregation": "individual"
 		    },
 		    "interfaces": [
 		      {
--- a/ci/vm.sh
+++ b/ci/vm.sh
@@ -245,6 +245,9 @@ teardown() {
 
 }
 
+# This function exec's the last binary run, this seems weird but actually works out ok.
+# Bash ends up running this function in subprocesses so the exec replace run_vm's process not ci/vm.sh's
+# This is needed so that run_vm callers will get qemu's pid and can kill it (otherwise bash won't forward the SIGTERM received by the function)
 run_vm() {
 	local cpu='' machine=''
 
@@ -281,7 +284,7 @@ run_vm() {
 	esac
 
 	# shellcheck disable=SC2068
-	"qemu-system-$arch" \
+	exec "qemu-system-$arch" \
 		"$@" \
 		-monitor unix:monitor.sock,server=on,wait=off \
 		-nographic \
--- a/ci/vm.sh
+++ b/ci/vm.sh
@@ -59,8 +59,7 @@ make_disk() {
 }
 
 gen_metadata() {
-	local class=$1 slug=$2 tag=$3
-	id=$(uuidgen)
+	local class=$1 slug=$2 tag=$3 id=$4
 
 	local cprcmd
 	cprcmd=(cat)
@@ -320,8 +319,9 @@ do_test() {
 
 	configure_nics
 
+	id=$(uuidgen)
 	# rename disk from scsi names to virtio names, e.g. sda1 -> vda1
-	gen_metadata "$class" "$slug" "$tag" <"$scriptdir/cpr/$class.cpr.json"
+	gen_metadata "$class" "$slug" "$tag" "$id" <"$scriptdir/cpr/$class.cpr.json"
 
 	start_dhcp \
 		--dhcp-host="${macs[0]},$pubip4" \
--- a/ci/vm.sh
+++ b/ci/vm.sh
@@ -200,10 +200,12 @@ start_web() {
 		}
 
 		metadata.packet.net:80 {
+		    browse
 		    log stderr
 		}
 
 		metadata.packet.net:443 {
+		    browse
 		    log stderr
 		    tls server.pem server-key.pem
 		}
--- a/ci/vm.sh
+++ b/ci/vm.sh
@@ -59,21 +59,19 @@ make_disk() {
 }
 
 gen_metadata() {
-	local cprcmd hardware_id id
+	local class=$1 slug=$2 tag=$3
 	id=$(uuidgen)
-	hardware_id=$(uuidgen)
-	local class=$1
-	local slug=$2
-	local tag=$3
-	local distro=${slug%%_*}
-	local version=${slug#*_}
-	version=${version//_/.}
 
+	local cprcmd
 	cprcmd=(cat)
-
 	if [[ $UEFI == true ]] && [[ $arch != aarch64 ]]; then
 		cprcmd=(jq -S '.filesystems += [{"mount":{"create":{"options": ["32", "-n", "EFI"]},"device":"/dev/sda1","format":"vfat","point":"/boot/efi"}}]')
 	fi
+
+	local distro=${slug%%_*} version=${slug#*_}
+	version=${version//_/.}
+	local hardware_id
+	hardware_id=$(uuidgen)
 
 	cat <<-EOF | sed '/\bsd[a-z]\+[0-9]*\b/ s|\bs\(d[a-z]\+[0-9]*\)\b|v\1|' | jq -S . | tee metadata
 		{
@@ -249,7 +247,7 @@ teardown() {
 }
 
 run_vm() {
-	local bios=() cpu='' machine=''
+	local cpu='' machine=''
 
 	case $(uname -m)-$arch in
 	'aarch64-aarch64') machine=virt cpu=host ;;
@@ -259,6 +257,7 @@ run_vm() {
 	*) echo 'unknown host-virt architecture combination' && exit 1 ;;
 	esac
 
+	local bios=()
 	if [[ $UEFI != 'true' ]]; then
 		bios=('-bios' '/usr/share/qemu/bios.bin')
 	else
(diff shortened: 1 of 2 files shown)


ci/Dockerfile
```diff
@@ -35,16 +35,14 @@ RUN apk add --no-cache --no-progress --update --upgrade \
         && \
     mv /etc/apk/repositories /etc/apk/repositories.old && \
     apk add --no-cache --update --upgrade \
-        --repository=http://dl-cdn.alpinelinux.org/alpine/v3.11/main \
-        --repository=http://dl-cdn.alpinelinux.org/alpine/v3.11/community \
+        --repository=http://dl-cdn.alpinelinux.org/alpine/v3.12/main \
```
tstromberg: This version bump appears to be an unrelated change and not mentioned in the PR description. 

* If necessary - add thy why to the PR description.
* If not necessary, move into a follow-up PR.

ci/Dockerfile
```diff
@@ -35,16 +35,14 @@ RUN apk add --no-cache --no-progress --update --upgrade \
...
+        --repository=http://dl-cdn.alpinelinux.org/alpine/v3.12/community \
         lshw \
         && \
-    mv /etc/apk/repositories.old /etc/apk/repositories && \
-    apk add --no-cache --update --upgrade \
-        --repository=http://dl-cdn.alpinelinux.org/alpine/edge/testing \
-        cfssl \
+    mv /etc/apk/repositories.old /etc/apk/repositories \
```
tstromberg: I'm not familiar with our style guidance here - but the line continuation here is unnecessary and the trailing semi-colon is highly irregular. Can both be removed?

ci/Dockerfile
```diff
@@ -35,16 +35,14 @@ RUN apk add --no-cache --no-progress --update --upgrade \
...
    ;
 
+RUN GOBIN=/bin GO111MODULE=on go get -v github.com/cloudflare/cfssl/cmd/cfssl@v1.6.0 github.com/cloudflare/cfssl/cmd/cfssljson@v1.6.0
```
tstromberg: Also not documented in the PR description - is it necessary?

ci/vm.sh
```diff
@@ -59,12 +59,12 @@ make_disk() {
 }
 
 gen_metadata() {
-	local cprcmd hardware_id id
-	id=$(uuidgen)
+	local cprcmd hardware_id
 	hardware_id=$(uuidgen)
```
tstromberg: I feel like handling the input variables should come first. 

hardware_id should be made a local, and moved closer to it's first usage.

ci/vm.sh
```diff
@@ -59,12 +59,12 @@ make_disk() {
...
 	local class=$1
 	local slug=$2
 	local tag=$3
+	local id=$4
```
tstromberg: This could use some disambiguation in naming: what kind of ID? is it a user id?

Alternatively, document the function inputs.

ci/vm.sh
```diff
@@ -283,7 +283,7 @@ run_vm() {
 	esac
 
 	# shellcheck disable=SC2068
-	"qemu-system-$arch" \
+	exec "qemu-system-$arch" \
```
tstromberg, line 296: Since `exec` from a non-main function is exotic and unexpected - can you add a  one-line comment to `run_vm()` describing that this function will never return?

ci/vm.sh
```diff
@@ -404,6 +410,27 @@ test_provision() {
 	diff -u <(jq -cS . <<<"$want") <(jq -cS . <<<"$got")
 }
 
+test_boot_and_phone_home() {
+	run_vm &
+	local vmpid=$! i=0 t=10
```
tstromberg: This variable declaration would be more readable if split across multiple statements.

ci/vm.sh
```diff
@@ -404,6 +410,27 @@ test_provision() {
...
+	until grep -qr instance_id uploads; do
+		sleep $t
```
tstromberg: just hardcode this as 10 for now and remove the variable.

markjacksonfishing (dismissed):
/lgtm
nothing looks off here

ci/vm.sh
```diff
@@ -404,6 +410,27 @@ test_provision() {
...
+		# timeout after 1min
+		if ((i++ == 6)); then
+			break
+		fi
+	done
+	echo 'system_powerdown' | socat unix-connect:monitor.sock -
+	timeout 30 sh -c "wait $vmpid" || kill -KILL $vmpid
+
+	# check for phone-home
+	eventid=instance_id
```
tstromberg: set local here and elsewhere.

ci/vm.sh
```diff
@@ -353,6 +354,11 @@ do_test() {
 	rm -f uploads/*
 
 	color=34
+	colorize $color "== Running Boot & Phone-Home Test =="
+	test_boot_and_phone_home |& stdbuf -i 0 sed "s/^/$(colorize $color 'test_boot_and_phone_home│')/" && echo "this test is expected to fail" >&2 && exit 1
```
tstromberg: break this statement out into multiple lines?

tstromberg (changes requested):

ci/Dockerfile
mmlb: will break out

ci/Dockerfile
mmlb: The trailing semi-colon enables easy copy&paste/deletion/moving and not needing to touch other lines. Along the same pattern as in go/python for trailing `,` in maps/dicts.

The trailing `;` is necessary to inform docker that the `RUN` is done, otherwise docker eats any empty lines and tries to run the next `COPY/RUN/...` as a shell script.

ci/Dockerfile
mmlb: Documented in the commit it came in from (https://github.com/tinkerbell/osie/pull/227/commits/84baf06ca2b63f987c30f583846c7dc5ff17ba12).

I'd rather keep the PR description with the important parts of the PR and leave any unrelated-yet-necessary commits documented solely in the commit message. I don't feel super strongly about that though, so if you'd like it in the PR description I can add it in.

ci/vm.sh
mmlb: > I feel like handling the input variables should come first.

Fair point, can fix.

> hardware_id should be made a local, and moved closer to it's first usage.

It is local, right there on line 62 ;). Going to split into 2 lines, and closer to use.

ci/vm.sh
mmlb: Will rename

ci/vm.sh
mmlb: It actually does return since its being called from within a function. I came on it by accident actually, I noticed that the `kill $vmpid` wasn't actually killing qemu and then noticed that qemu and the backgrounded function had different pids (so that explains the kill issue).

ci/vm.sh
mmlb: :thinking: I need to double check this because I thought I needed a different branch for non-UEFI that was supposed to actually work.

ci/vm.sh
mmlb: I'll just get rid of `i` and `t`, they were useful when I was first running this test but no longer really relevant.

ci/vm.sh
mmlb: :+1:

ci/vm.sh
mmlb: :+1:

ci/vm.sh
mmlb: Which isn't that weird when you think about it. Bash/sh already runs the function in a subshell if there's any external calls so the `exec` just takes over the func's process.

```ShellSession
12:33  ~
$ declare -f f
f () 
{ 
    echo f:ppid:$$;
    g
}

12:33  ~
$ declare -f g
g () 
{ 
    /usr/bin/env bash -c 'echo g:ppid:$$'
}

12:33  ~
$ echo $$
64465

12:33  ~
$ f
f:ppid:64465
g:ppid:64802
```

mmlb: Actually nm, my example isn't actually relevant :D.

mmlb: I'm not sure exactly *what* causes bash to run `run_vm` in a subprocess but it sure does and the exec takes over that sub process. I'll add a little bit of docs to the func and maybe change from:

```bash
run_vm() { ... }
```
to 
```bash
run_vm() ( ... )
```
to ensure its run in a subprocess

ci/Dockerfile
mmlb: removed

ci/vm.sh
mmlb: Fixed

ci/vm.sh
mmlb: fixed

ci/vm.sh
mmlb: Added some comments to the function and changed to `run_vm() (...)` form. Better?

ci/vm.sh
mmlb: Glad you called me out on this because I had some other stuff laying around in a stash that I actually wanted to have in the PR :D.

markjacksonfishing (dismissed):

markyjackson-taulia:
@mmlb for PR that is WIP, can you set them in the draft so it can get a review when it is ready? This will help cut down on reviews when the PR is actually not ready

mmlb:
oooooookay! tests are green again. I've rolled back the "test all the oses/arches/fws" commits for another PR so we can get this one out and done with! PTAL again.

mmlb:
@tstromberg PTAL. (I've mashed the "re-request review" button a dozen times but it doesn't seem to change state, I hope I didn't spam you)

ci/vm.sh
```diff
@@ -59,22 +59,19 @@ make_disk() {
...
+	local class=$1 slug=$2 tag=$3 id=$4
 
+	local cprcmd
 	cprcmd=(cat)
-
 	if [[ $UEFI == true ]] && [[ $arch != aarch64 ]]; then
 		cprcmd=(jq -S '.filesystems += [{"mount":{"create":{"options": ["32", "-n", "EFI"]},"device":"/dev/sda1","format":"vfat","point":"/boot/efi"}}]')
 	fi
 
+	local distro=${slug%%_*} version=${slug#*_}
```
tstromberg, line 70: optional nit: prefer declaration and assignment as separate statements. 

There is no behavioral difference here, but it prevents awkward behaviors when used with command substitution; as the local builtin does not propagate the exit code from the command substitution.

Reference: https://google.github.io/styleguide/shellguide.html#use-local-variables

(feel free to disregard this comment as it seems to disagree with this code base)

ci/vm.sh
```diff
@@ -144,6 +141,8 @@ gen_metadata() {
 		  "wipe_disks": true
 		}
 	EOF
+	mkdir -p 2009-04-04/meta-data
```
tstromberg, line 147: Please document this magic value (2009-04-04) or set it to a variable with a describing string. I imagine that this value has to be identical across a couple of places here.

ci/Dockerfile
tstromberg: When reviewing a PR, it should only be necessary to read the PR description to get the necessary context. Not only does this let the reviewer know which commits were intentional versus accidental, but one cannot expect that a reviewer is going to trawl through all of the commit messages.

My expectation is that PR descriptions contain roughly:
- What is being done
- Why is it being done?
- Why is this the best approach to take?

ci/vm.sh
```diff
@@ -353,6 +361,20 @@ do_test() {
 	rm -f uploads/*
 
 	color=34
+	colorize $color "== Running Boot & Phone-Home Test =="
+	if [[ $arch == aarch64 ]]; then
+		# aarch64 doesn't work, idk why
```
tstromberg, line 369: Please add a TODO with an issue URL to track this problem.

docker/scripts/osie.sh
```diff
@@ -370,7 +370,10 @@ EOF
 			  Ec2:
 			    timeout: 60
 			    max_wait: 120
-			    metadata_urls: [ 'http://metadata.packet.net', 'http://147.75.207.1:80' ]
+			    metadata_urls:
```
tstromberg, line 373: The empty line before the list of URL's is suspicious -- is it intentional?

tstromberg (changes requested):

tstromberg (dismissed):

ci/vm.sh
mmlb: Shellcheck errors on the command substitution case. I tend to do the obvious/simplest/least-lines case unless a tool complains. In this case:
```bash
local foo=1 bar=2
```
seems more obvious than:
```bash
local foo bar
foo=1
bar=2
```
The latter would have me wondering why its separated. I think for shellscripts its kind of weird as most uses are interactive and the latter case (separate declaration and definition) is something anyone will do at a command prompt.

docker/scripts/osie.sh
mmlb: metadata_urls is an array. I've changed from `[]` syntax to `lines prefixed with -` instead.

ci/vm.sh
mmlb: ^ is *not* something anyone would do at a prompt

ci/vm.sh
mmlb: done

ci/vm.sh
mmlb: done

mmlb:
Hey @tstromberg addressed some of your comments. Dropped the cfssl commit as it seems it missing from mirrors was temporary and now available again, noted the PR description details though.

tstromberg (approved):

PR #231, opened 2021-06-18
Help OSIE talk to S3

ScottGarman:
Setting OSIE up to be able to talk to S3 interface, to allow for eventual migration to a s3-based OS images.


docker/get-package-list.sh
```diff
@@ -56,6 +56,7 @@ packages=(
 	pciutils
 	pv
 	python3
+	s3cmd
```
andy-v-h: we're using `curl` now

ScottGarman: I've now added a commit to remove this.

markjacksonfishing (approved):

PR #235, opened 2021-06-21
apps/workflow-helper: set CAPTURE_ACTION_LOGS to false in tink-worker env

splaspood:
## Description

Adds CAPTURE_ACTION_LOGS=false to the tink-worker env as defined in apps/workflow-helper.sh.    This prevents tink-worker from 'attaching' to the log output of each action container it spawns.   

## Why is this needed

To send all workflow action output to a centralized log repository by way of boots we need to prevent tink-worker from grabbing that log output first.


markjacksonfishing (dismissed):

mmlb (changes requested):
Needs a more descriptive commit message, code looks fine though.

ScottGarman (approved):

mmlb (approved):

PR #240, opened 2021-06-28
make: Modify the "deploy this build..." message

mmlb:
## Description

Changes the git based version scheme to use just the output of `git describe --dirty` since we don't really need all the other info from before and the format is not allowed as docker image tags.

This PR also truncates the commit sha injected by drone to be just the first 8 hex characters, no need for the full commit id and is much easier to interact with.

Also updating the post deploy-to-s3 message about deploying to production to get rid of EM specific stuff.

## Why is this needed

Local tests fail because the docker osie-test-env image tag that now uses the version info uses forbidden characters (`=` and `,`).

## How Has This Been Tested?

Locally built and CI.

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/rules.mk.j2
+++ b/rules.mk.j2
@@ -36,23 +36,6 @@ endif
 
 ifndef v
 v := $(shell git describe --dirty)
-ifeq (-g,$(findstring -g,$v))
-build_git_version := true
-endif
-ifeq (-dirty,$(findstring -dirty,$v))
-build_git_version := true
-endif
-
-ifeq (true,${build_git_version})
-t := $(word 1, $(subst -, ,$v))
-n := $(word 2, $(subst -, ,$v))
-c := $(shell git rev-parse --short HEAD)
-d := ,dirty
-d := $(subst dirty,$d,$(findstring dirty,$v))
-b := $(shell git symbolic-ref HEAD | sed -e 's|^refs/heads/||' -e 's|/|-|g' -e 's|^|,b=|')
-v := $t-n=$n,c=$c$b$d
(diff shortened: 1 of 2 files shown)


rules.mk.j2
```diff
@@ -118,11 +101,9 @@ deploy-to-s3: package
 		sed 's| (.*) | (latest.tar.gz) |' build/osie-$v.tar.gz.sha512sum >build/latest.tar.gz.sha512sum
 		mc cp s3/tinkerbell-oss/osie-uploads/latest.tar.gz.sha512sum s3/tinkerbell-oss/osie-uploads/latest.tar.gz.sha512sum
 	fi
-	echo "deploy this build from the deploy-osie repo with the following command:"
-	echo -n "./deploy update osie-$v -m \"$$"
-	echo -n "(sed -n '/^## \[/,$$ {/\S/!q; p}' CHANGELOG.md)\" "
-	echo -n "$$(sed 's|.*= ||' build/osie-$v.tar.gz.sha512sum)"
-	echo
```
ScottGarman: I regularly rely on this info when deploying osie, and I know I've taught several other folks to deploy osie using this info as reference. I realize it's EM specific but I'd really like to keep that to avoid further hassles.

mmlb: yeah I was on the fence about it, let me re-word it instead.

mmlb: I've updated the messaging a bit more, PTAL again.

ScottGarman (approved):

Issue #245, opened 2021-08-04
osie-runner triggering SIGSEGV and being terminated


nshalman:
We've seen some machines where preinstallation has completed and osie-runner is sitting connected to hegel over grpc where it suddenly terminates with exit code 139:

```
localhost:~# docker ps -a
CONTAINER ID        IMAGE                COMMAND                  CREATED             STATUS                     PORTS               NAMES
bcc4cb56c565        osie-runner:x86_64   "/entrypoint.sh pyth…"   5 days ago          Exited (139) 7 hours ago                       quizzical_varahamihira
localhost:~# docker logs quizzical_varahamihira --tail 3
}
2021-08-04T10:40:45.952178Z [info     ] no handler for state           [runner] state=provisionable
2021-08-04T10:40:45.952249Z [info     ] about to monitor               [runner]
```

## Expected Behaviour
osie-runner keeps running indefinitely until it receives new information from hegel

## Current Behaviour
osie-runner exits unexpectedly with exit code 139

## Possible Solution
Update to the latest grpc in hopes that some bug in the core library has been fixed.

## Steps to Reproduce (for bugs)
1.
2.
3.
4.

## Your Environment
Equinix Metal Production

nshalman:
Closing as hopefully addressed by #246. Will re-open if we see it again.

PR #244, opened 2021-07-29
Enable grpc keepalive in osie-runner

nshalman:
## Description

Enable grpc keepalive in osie-runner
See https://cs.mcgill.ca/~mxia3/2019/02/23/Using-gRPC-in-Production/

## Why is this needed

osie-runner currently doesn't notice unclean network disconnects

## How Has This Been Tested?

Testing is in progress.

## How are existing users impacted? What migration steps/scripts do we need?

osie-runner would need to be deployed

## Checklist:

I have:

- [ ] updated the documentation and/or roadmap (if required)
- [ ] added unit or e2e tests
- [ ] provided instructions on how to upgrade

PR #246, opened 2021-08-04
osie-runner: Update python packages

nshalman:
## Description

Update the python packages for osie-runner to pull in the latest grpc.

## Why is this needed

With the merging of #244 we started seeing the issue documented in #245 and I'm hoping it's a bug in the grpc core that has since been fixed.

Relates to: #245 

## How Has This Been Tested?
- [x] Does it build?
- [x] Does it run?
- [ ] Test in production

If anyone else is using osie-runner in production, they might be impacted by #245  and would need to upgrade to the newest version after this fix lands if this indeed fixes the problem

## Checklist:

I have:

- [ ] added unit or e2e tests


ScottGarman (approved):
Approved assuming CI passes.

PR #248, opened 2021-08-05
Write json output from packet-hardware inventorybios to Hollow

ScottGarman:
Did this in such a way that errors should fail open and log details, rather than failing the deprovision.

### DIFF
--- a/docker/scripts/functions.sh
+++ b/docker/scripts/functions.sh
@@ -403,6 +403,78 @@ function validate_bios_config() {
 	fi
 }
 
+# usage: bios_inventory $id $plan $facility
+function bios_inventory() {
+	local id=$1
+	local plan=$2
+	local facility=$3
+
+	# When running the inventorybios command outside of the packet-hardware
+	# container, UTIL_RACADM7 and UTIL_SUM must be set to the locations of the
+	# racadm and sum binaries, respectively
+	local bios_json_fn="/tmp/bios.json"
+	local hollow_json_fn="/tmp/hollow.json"
+	local hollow_namespace="net.equinixplatform.bios"
+	if UTIL_RACADM7=/opt/dell/srvadmin/bin/idracadm7 UTIL_SUM=/opt/supermicro/sum/sum packet-hardware inventorybios --verbose -u localhost --dry --cache-file "${bios_json_fn}"; then
+		# Generate JSON with additional fields required by Hollow
+		if ! jq --null-input \
+			--arg ns "${hollow_namespace}" \
+			--arg id "${id}" \
+			--slurpfile data "${bios_json_fn}" \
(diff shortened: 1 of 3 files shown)


docker/scripts/functions.sh
```diff
@@ -403,6 +403,81 @@ function validate_bios_config() {
...
+
+		# Write data to Hollow
+		if [[ -z ${HOLLOW_CLIENT_ID:-} ]] || [[ -z ${HOLLOW_CLIENT_REQUEST_SECRET:-} ]]; then
+			echo "Warning: env credentials for Hollow not found, not writing to Hollow"
+			set -x
+			return 0
+		fi
+
+		hollow_auth_url="https://auth.equinixmetal.com/oauth/token"
+		hollow_audience="https://hollow.platformequinix.net"
```
mmlb: why not declare	`local hollow_auth_url="..."` here instead of old-school-C-like split?
It would be nice to make these `local -r` for readonly, even though its not common in this code base but no time like the present to start! (or not)

mmlb (dismissed):

docker/scripts/functions.sh
ScottGarman: I've moved the local declarations down where shellcheck doesn't complain about them (`hollow_token` and `hollow_response` are not included as shellcheck complains that the local declaration might mask a return value).

I'm leaving off the `local -r` declarations, and feel it's better to do this consistently in one pass across the script(s) at a future time.

docker/scripts/functions.sh
```diff
@@ -403,6 +403,78 @@ function validate_bios_config() {
...
+			echo "Warning: env credentials for Hollow not found, not writing to Hollow"
+			set -x
+			return 0
+		fi
+
+		local hollow_auth_url="https://auth.equinixmetal.com/oauth/token"
+		local hollow_audience="https://hollow.platformequinix.net"
+		# Use eclypsium-proxy to reach the auth server from the deprov network
+		echo "Requesting a hollow token from ${hollow_auth_url}"
+		if ! hollow_token=$(HTTPS_PROXY="http://eclypsium-proxy-${facility}.packet.net:8888/" curl --request POST \
```
mmlb: If you change to
```suggestion
		echo "Requesting a hollow token from ${hollow_auth_url}"
		local hollow_token
		if ! hollow_token=$(HTTPS_PROXY="http://eclypsium-proxy-${facility}.packet.net:8888/" curl --request POST \
```
you can bring it down to here, ditto with hollow_url

mmlb (changes requested):

docker/scripts/functions.sh
ScottGarman: Done.

mmlb (approved):

PR #249, opened 2021-08-06
deprovision: add support for PERC H745

dustinmiller:
Signed-off-by: Dustin Miller <dustin@packet.com>

## Description

Lets do a perc_reset on the Dell PERC H745

## Why is this needed

Seems currently we're not matching this for the perc_reset function
and we get snared trying to run some megacli64 commands.

```
+ MegaCli64 -AdpSetProp -EnableJBOD -1 -a0"
-\r                                     "
Adapter 0: Failed to Set Adapter Properties."
```

## How Has This Been Tested?

Via custom OSIE deprovisions

### DIFF
--- a/docker/scripts/functions.sh
+++ b/docker/scripts/functions.sh
@@ -800,7 +800,7 @@ function perc_reset() {
 		perccli64 /c0 set personality=HBA
 	fi
 
-	#Check/delete all VDs! -- VDs and VD's it seems
+	#Check/delete all VDs!
 	if perccli64 /c0 /vall show | grep "No VD" >/dev/null; then
 		echo "PERCCLI - No VDs configured - OK"
 	else
--- a/docker/scripts/functions.sh
+++ b/docker/scripts/functions.sh
@@ -800,8 +800,8 @@ function perc_reset() {
 		perccli64 /c0 set personality=HBA
 	fi
 
-	#Check/delete all VDs!
-	if perccli64 /c0 /vall show | grep "No VDs" >/dev/null; then
+	#Check/delete all VDs! -- VDs and VD's it seems
+	if perccli64 /c0 /vall show | grep "No VD" >/dev/null; then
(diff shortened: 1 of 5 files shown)


dustinmiller:
Not sure if we need newer perccli or not, just a guess there. 

Here's how it failed using perc_reset

```
perccli64 /c0 set personality=HBA"
 Controller = 0"
 Status = Failure"
 Description = None"
 "
 Detailed Status :"
 ==============="
 "
 --------------------------------------------------------------"
 Ctrl Status Ctrl_Prop       Value ErrMsg                ErrCd "
 --------------------------------------------------------------"
    0 Failed Set Personality -     setPersonality failed     3 "
 --------------------------------------------------------------"
 "
 "
 + autofail"
 + (( 3 == 0 ))" 
```

docker/scripts/functions.sh
```diff
@@ -788,17 +788,20 @@ function perc_reset() {
...
+		echo "PERCCLI - Setting personality to eHBA-Mode"
+		perccli64 /c0 set personality=eHBA
 	else
 		echo "PERCCLI - Setting personality to HBA-Mode"
 		perccli64 /c0 set personality=HBA
 	fi
 
-	#Check/delete all VDs!
-	if perccli64 /c0 /vall show | grep "No VDs" >/dev/null; then
+	#Check/delete all VDs! -- VDs and VD's it seems
```
tstromberg: This comment is going to seem very cryptic to someone unless they dig into the related PR. Perhaps keep it as is?

tstromberg (dismissed):
Approved with one minor change request

docker/scripts/functions.sh
dustinmiller: https://github.com/tinkerbell/osie/pull/249/commits/f69d8922a8957c2a096a723845f96abaebc1a011 reverts to the original for clarity

tstromberg (approved):

PR #250, opened 2021-08-09
Write json output from packet-hardware inventorybios to Hollow

ScottGarman:
Did this in such a way that errors should fail open and log details, rather than failing the deprovision.


ScottGarman:
Closing, this was a dupe.

PR #253, opened 2021-08-13
Avoid sending empty BIOS invetories to Hollow

ScottGarman:
The 'packet-hardware inventorybios' command can succeed but generate
empty BIOS JSON on unsupported platforms. Check for this and avoid
writing to Hollow in these cases.

The `$id` that we get from metadata is the instance uuid, not the hardware uuid. Make the hwuuid available to osie scripts by passing it down via osie runner's docker env.

Also fixed a typo in the hollow namespace.

### DIFF
--- a/docker/scripts/functions.sh
+++ b/docker/scripts/functions.sh
@@ -403,9 +403,9 @@ function validate_bios_config() {
 	fi
 }
 
-# usage: bios_inventory $id $plan $facility
+# usage: bios_inventory $hwuuid $plan $facility
 function bios_inventory() {
-	local id=$1
+	local hwuuid=$1
 	local plan=$2
 	local facility=$3
 
@@ -416,10 +416,19 @@ function bios_inventory() {
 	local hollow_json_fn="/tmp/hollow.json"
 	local hollow_namespace="net.equinixplatform.bios"
 	if UTIL_RACADM7=/opt/dell/srvadmin/bin/idracadm7 UTIL_SUM=/opt/supermicro/sum/sum packet-hardware inventorybios --verbose -u localhost --dry --cache-file "${bios_json_fn}"; then
+		local inventorybios_json
+		inventorybios_json="$(cat "${bios_json_fn}")"
+		echo "inventorybios JSON is: [${inventorybios_json}]"
+
+		if [[ ${inventorybios_json} == "{}" ]]; then
+			echo "inventorybios JSON is empty, not writing to Hollow"
+			return 0
+		fi
+
 		# Generate JSON with additional fields required by Hollow
 		if ! jq --null-input \
 			--arg ns "${hollow_namespace}" \
-			--arg id "${id}" \
+			--arg id "${hwuuid}" \
 			--slurpfile data "${bios_json_fn}" \
 			'{Namespace: $ns, hardware_uuid: $id, data: $data[]}' >"${hollow_json_fn}"; then
 			echo "Warning: error while generating hollow json"
@@ -453,7 +462,7 @@ function bios_inventory() {
 			return 0
 		fi
 
-		local hollow_url="https://hollow.edge-a.${facility}.metalkube.net/api/v1/servers/${id}/versioned-attributes"
+		local hollow_url="https://hollow.edge-a.${facility}.metalkube.net/api/v1/servers/${hwuuid}/versioned-attributes"
 		# Write the bios data to Hollow
 		echo "Writing BIOS feature inventory data to ${hollow_url}"
 		local hollow_response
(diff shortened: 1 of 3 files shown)


docker/scripts/functions.sh
```diff
@@ -403,23 +403,31 @@ function validate_bios_config() {
...
+	local plan=$1
+	local facility=$2
 
 	# When running the inventorybios command outside of the packet-hardware
 	# container, UTIL_RACADM7 and UTIL_SUM must be set to the locations of the
 	# racadm and sum binaries, respectively
 	local bios_json_fn="/tmp/bios.json"
 	local hollow_json_fn="/tmp/hollow.json"
-	local hollow_namespace="net.equinixplatform.bios"
+	local hollow_namespace="net.platformequinix.bios"
```
tstromberg, line 417: Please mention this change in the PR description and/or move it into another PR.

docker/scripts/functions.sh
```diff
@@ -403,23 +403,31 @@ function validate_bios_config() {
...
 	if UTIL_RACADM7=/opt/dell/srvadmin/bin/idracadm7 UTIL_SUM=/opt/supermicro/sum/sum packet-hardware inventorybios --verbose -u localhost --dry --cache-file "${bios_json_fn}"; then
+		local inventorybios_json
+		inventorybios_json=$(cat ${bios_json_fn})
```
tstromberg: Always quote command substitutions (unless you need unquoted behavior): https://google.github.io/styleguide/shellguide.html#quoting

docker/scripts/functions.sh
```diff
@@ -403,23 +403,31 @@ function validate_bios_config() {
...
+		echo "inventorybios JSON is: [${inventorybios_json}]"
+
+		if [[ ${inventorybios_json} == "{}" ]]; then
```
tstromberg, line 423: you should add quotes here too, otherwise certain strings ("", "|") may act funny.

docker/scripts/functions.sh
```diff
@@ -403,23 +403,31 @@ function validate_bios_config() {
...
+			echo "inventorybios JSON is empty, not writing to Hollow"
+			return 0
+		fi
+
 		# Generate JSON with additional fields required by Hollow
 		if ! jq --null-input \
 			--arg ns "${hollow_namespace}" \
-			--arg id "${id}" \
+			--arg id "${HARDWARE_ID}" \
```
tstromberg: It feels really weird to pass this in as a global environment variable, when the other inputs are arguments. I'd recommend consistency. 

If you do need to rely on global variables passed-in from elsewhere - please document them as as a doc string in the function signature.

tstromberg (changes requested):

docker/scripts/functions.sh
ScottGarman: This change was done in a separate commit, but I failed to mention it in the PR description. That's been fixed, and I've improved the PR description to be more complete in other ways, too.

docker/scripts/functions.sh
ScottGarman: Fixed.

docker/scripts/functions.sh
ScottGarman: Fixed (I keep thinking shellcheck will catch these, but it doesn't).

docker/scripts/functions.sh
ScottGarman: The reason for this change wasn't mentioned in the commit message or PR description - I've updated both for clarity as to why. I agree that it's clearer to pass the value is a function parameter, so I've made that change as well.

docker/scripts/functions.sh
ScottGarman: Actually, I misspoke. It's `shfmt` which doesn't like those quotes, so I have to leave them off in order to pass CI. This doesn't seem right to me either, and a quick search online doesn't reveal the reasoning behind this - I may have to dig in further.

mmlb: the quotes issues is *only* for `sh`'s `[` and not for `bash`'s `[[ ]]` which is builtin:

```ShellSession
14:52  ~
$ foo='"'

14:52  ~
$ [[ $foo == '' ]] && echo true

14:52  ~
$ [[ $foo == '|' ]] && echo true

14:52  ~
$ [[ $foo == '"' ]] && echo true
true

14:52  ~
$ foo='|'

14:52  ~
$ [[ $foo == '' ]] && echo true

14:52  ~
$ [[ $foo == '|' ]] && echo true
true

14:52  ~
$ [[ $foo == '"' ]] && echo true
```

tstromberg (approved):

PR #262, opened 2021-08-26
detect vmx/svm and fall back to qemu software mode

tobert:
## Description

Looks for the vmx/svm cpu flags in /proc/cpuinfo and falls back to qemu software mode when they're not present.

## Why is this needed

Makes it possible to run the VM test on a VM.

I do most of my work in an Xhyve/hyperkit VM on my Mac. This change makes it so the VM will start up in my dev environment and probably makes it more readily available to others.

Should be no change for most folks.

I have not tested on bare metal yet as I don't have a machine set up right now.

### DIFF
--- a/ci/vm.sh
+++ b/ci/vm.sh
@@ -268,7 +268,7 @@ run_vm() {
 
 	# fall back to qemu software mode when KVM isn't available
 	local kvm_accel=",accel=kvm:tcg"
-	if ! (grep -Eq ' (svm|vmx)' /proc/cpuinfo) ; then
+	if ! (grep -Eq ' (svm|vmx)' /proc/cpuinfo); then
 		cpu=qemu64
 		kvm_accel=""
 	fi
@@ -305,7 +305,7 @@ run_vm() {
 		-monitor unix:monitor.sock,server=on,wait=off \
 		-nographic \
 		"${serials[@]}" \
-		-machine $machine$kvm_accel -cpu $cpu -smp 2 -m 8192 \
+		-machine "$machine$kvm_accel" -cpu $cpu -smp 2 -m 8192 \
 		-drive if=virtio,file="$disk",format=raw,discard=unmap \
 		-object rng-random,filename=/dev/urandom,id=rng0 \
 		-device virtio-rng-pci,rng=rng0 \
--- a/ci/vm.sh
+++ b/ci/vm.sh
@@ -257,21 +257,22 @@ teardown() {
 # This is needed so that run_vm callers will get qemu's pid and can kill it (otherwise bash won't forward the SIGTERM received by the function)
 run_vm() {
 	local cpu='' machine=''
+	local kvm_accel=',accel=kvm:tcg'
 
 	case $(uname -m)-$arch in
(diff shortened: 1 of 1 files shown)


tobert:
I have no idea how this build failure is related to my change?

ScottGarman:
> I have no idea how this build failure is related to my change?

That was an odd one. I simply restarted the build and it passed this time.

tobert:
Oh yeah, do we use KVM on ARM at all? My check for vmx/svx probably won't do the right thing there.

mmlb:
> Oh yeah, do we use KVM on ARM at all? My check for vmx/svx probably won't do the right thing there.

Nope we don't. I've never tried building OSIE on arm. Once linux works a bit better I'm planning on grabbing a M1 mac mini and we'll find out! This is fine for now :D.

ci/vm.sh
```diff
@@ -266,6 +266,13 @@ run_vm() {
 	*) echo 'unknown host-virt architecture combination' && exit 1 ;;
 	esac
 
+	# fall back to qemu software mode when KVM isn't available
+	local kvm_accel=",accel=kvm:tcg"
+	if ! (grep -Eq ' (svm|vmx)' /proc/cpuinfo); then
+		cpu=qemu64
```
mmlb: do we really need to override the cpu?

mmlb:
@tobert sorry I didn't notice the cpu= thing and thought this was ready to go so rebased on top of master to make GH happy. So if we don't need the cpu= and you get rid of it you'll need to fetch/rebase.

tobert:
@mmlb I can't merge to these repos, so please merge at your leisure :)

ci/vm.sh
mmlb: @tobert this is still an outstanding question I've got. Maybe only do this if `machine=pc`

tobert: yeah, I tried just eliminating the accel tag first and that failed, so I added the change to qemu64.

tobert: with cpu=host it returns `qemu-system-x86_64: CPU model 'host' requires KVM` so that's why qemu64

tobert: refactor incoming that will make this a lil cleaner as you suggest

tobert:
@mmlb ok I refactored it a bit to only fallback on PC and validated that we do need to set cpu=qemu64 because cpu=host is only valid when KVM is available.

tstromberg (approved):

ci/vm.sh
```diff
@@ -298,7 +306,7 @@ run_vm() {
 		-monitor unix:monitor.sock,server=on,wait=off \
 		-nographic \
 		"${serials[@]}" \
-		-machine $machine,accel=kvm:tcg -cpu $cpu -smp 2 -m 8192 \
+		-machine "$machine$kvm_accel" -cpu $cpu -smp 2 -m 8192 \
```
tstromberg, line 309: Prefer "${machine}${kvm_accel}", particularly when variables get smashed up into each other: https://google.github.io/styleguide/shellguide.html#variable-expansion

I recognize that this file is all over the place for braces versus not braces.

tstromberg (approved):

PR #264, opened 2021-08-31
Remove unnecessary plan check during mdadm config update.

Raj-Dharwadkar:
## Description
 Plan check was not required because we see the same behavior for all plans.

## Why is this needed

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/docker/scripts/osie.sh
+++ b/docker/scripts/osie.sh
@@ -285,10 +285,8 @@ if ! [[ -f /statedir/disks-partioned-image-extracted ]]; then
 	phone_home "${tinkerbell}" '{"type":"provisioning.106"}'
 
 	mkdir -p $target/etc/mdadm
-	if [[ $class != "t1.small.x86" ]]; then
-		echo -e "${GREEN}#### Updating MD RAID config file ${NC}"
-		mdadm --examine --scan >>$target/etc/mdadm/mdadm.conf
-	fi
+	echo -e "${GREEN}#### Updating MD RAID config file ${NC}"
+	mdadm --examine --scan >>$target/etc/mdadm/mdadm.conf
 
 	# ensure unique dbus/systemd machine-id
 	set_autofail_stage "machine-id setup"


docker/scripts/osie.sh
```diff
@@ -285,7 +285,12 @@ if ! [[ -f /statedir/disks-partioned-image-extracted ]]; then
 	phone_home "${tinkerbell}" '{"type":"provisioning.106"}'
 
 	mkdir -p $target/etc/mdadm
-	if [[ $class != "t1.small.x86" ]]; then
+	if [[ $class == "c1.small.x86" ]]; then
+		echo -e "${GREEN}#### Updating MD RAID config file ${NC}"
+		mdadm --examine --scan >>$target/etc/mdadm/mdadm.conf
+	fi
+
+	if [[ $class != "c1.small.x86" ]]; then
```
mmlb: this change looks wrong, you got rid of t1.small.x86 and ended up calling `mdadm --examine ...` for both c1.small and !c1.small.

In any case, this is the wrong way to handle this I think. Instead we should add a boolean (has_raid) or int (raid_count) to https://github.com/tinkerbell/osie/blob/master/docker/scripts/cpr.sh#L164-L172 and then just check if the boolean exists or count >0 and then and only then add the `mdadm --examine`. This we also get rid of any plan/class name reference.

ScottGarman: I'd like to get more context on why this was added to osie.sh in the first place if you know the history. It seems that we run that mdadm command on every other plan type regardless of whether or not it has raid enabled or not, and if it does, I'd hope that all of that logic would be contained within the cpr.sh script. When running some tests, there seems to be no odd behavior of mdadm on t1.smalls that I have found. So my theory is that we should just be able to remove the if-based exclusion altogether, and leave cpr.sh as-is.

ScottGarman:
I think the code change is good, but the commit message should be changed. I'd recommend something like "Remove unnecessary plan reference during mdadm config"

ScottGarman (approved):

PR #266, opened 2021-09-01
add opentelemetry

tobert:
## Description

Implements OpenTelemetry tracing in OSIE. This includes some code, otel-cli spans in shell scripts, and W3C traceparent propagation in http headers where we can.

## Why is this needed

When everything is in place, we will get detailed traces of operations happening in OSIE, making it observable.

Understanding what is happening in complex provisioning systems is difficult. We mostly get the job done with logs and metrics these days. This is an attempt to add OpenTelemetry events so that users of OSIE and other Tinkerbell components can emit traces to their favorite OpenTelemetry-compatible provider (or translate through opentelemetry-collector).

## Checklist:

My TODO:

- [x] add otel-cli to the image build process
- [x] pass TRACEPARENT environment variables through docker calls
- [x] add W3C traceparent headers to curl calls in scripts
- [x] add opentelemetry-python to osie-runner (or at least propagate traceparent headers)
- [x] figure out the best way to configure otel-cli at runtime (propagate OTEL_EXPORTER_OTLP_*)

Related work:

- [ ] add TRACEPARENT envvar support to `tink` (WIP)
- [ ] teach boots http to accept traceparent as a query param (TBD)
- [x] merge boots PR to add tp in kernel command line
- [x] finish up otel-cli status support (https://github.com/equinix-labs/otel-cli/pull/27) (handy for reporting failures)
- [x] make `otel-cli span background --tp-carrier` write out a file that can be sourced
- [x] add `otel-cli span background --wait`

I have:

- [ ] added unit or e2e tests


osie-runner/run.py
```diff
@@ -102,6 +138,9 @@ def sanitize_cacher_data(j):
     cmdline_content = cmdline.read()
     tinkerbell = parse.urlparse(util.value_from_kopt(cmdline_content, "tinkerbell"))
     facility = util.value_from_kopt(cmdline_content, "facility")
+    kernel_tp = util.value_from_kopt(cmdline_content, "traceparent") # opentelemetry
```
splaspood: Wondering if maybe we want to shorten this kopt.  I keep going back and forth on it (it's not really a big deal either way) and maybe this comment is better made on the corresponding boots PR.

tobert: I seriously considered that, but given how new otel is to most folks, let alone those of us in bare metal land, I figured it was better to keep the long name for clarity when people ask "wtf is this?"

jacobweinstock:
Hey @ScottGarman , @splaspood would either of you be available, by chance, to check this out?

ScottGarman:
I'm :+1: on the overall gist of this, though I do wonder if it would be simpler to start with a specific component (like osie-runner) to instrument first, then roll it out more widely. Other than that, just request typical commit cleanup/squashes and a rebase to the main branch. I can then assist with the process of deploying an osie test build so we can run this in production at a limited scale.

tobert:
@ScottGarman I have been thinking about splitting this PR and might just do that. Getting the otel-cli bits to work right will need more work to pass around `OTEL_EXPORTER_OTLP_ENDPOINT` or some other configuration pointing at a server, so I think we can defer the otel-cli bits a while and see how things look after we light all this up.

tobert:
I know y'all don't like to leave these hanging. Gonna close this and will open new PRs soon with the same code split up.

PR #268, opened 2021-09-03
Check if nvme code runs on x.large.arm server.

Raj-Dharwadkar:
## Description

  Delete nvme namespace does not work servers(CPU manufacturer: Ampere and version: EVT2 or DVT or 0100), so
  skipping nvme namespace management on them.

  Delete storage detection because it is not required in deprovision.

## Why is this needed

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests


docker/scripts/deprovision.sh
```diff
@@ -149,8 +132,9 @@ if [[ $preserve_data == false ]]; then
 	nvme_drives=($(find /dev -regex ".*/nvme[0-9]+" | sort -h))
 	echo "Found ${#nvme_drives[@]} nvme drives"
 	nvme list
-	if [[ $class == "x.large.arm" ]]; then
-		echo "Skipping NVMe namespace management for $class hardware"
+	system_version=$(dmidecode --string system-version)
+	if [[ $system_version == "EVT2" ]]; then
+		echo "Skipping NVMe namespace management for $system_version system version"
```
ScottGarman, line 142: I would add one comment in here so that we can remember why this was added. Maybe something like "Deleting NVMe namespaces fails on these servers, so exclude them from NVMe namespace management"

Also, we are not only excluding the EVT2 servers - we also need to exclude this on the Lenovo HR350A models. So you should also check if `system_version` equals "HR350A".

Raj-Dharwadkar: Fixed.

Raj-Dharwadkar: > I would add one comment in here so that we can remember why this was added. Maybe something like "Deleting NVMe namespaces fails on these servers, so exclude them from NVMe namespace management"
> 
> Also, we are not only excluding the EVT2 servers - we also need to exclude this on the Lenovo HR350A models. So you should also check if `system_version` equals "HR350A".

Fixed.

ScottGarman (approved):

PR #284, opened 2021-09-22
Include the asrockrack bios kernel module 

joelrebel:
## Description

This change updates the alpine image to build and include the `asrdev.ko` kernel module,
which enables the Asrockrack BIOS utility `BIOSControl` to set/get BIOS configuration.

## Why is this needed

To export and import BIOS configuration on the Asrockrack hardware.

## How Has This Been Tested?

By building the Alpine docker image


mmlb:
Hey @joelrebel the zip file should be added to lfs, not directly to git do you have `git lfs` installed and enabled? You'll need to do something like `git lfs track installer/alpine/*.zip` so the git magic can happen.

ScottGarman:
Also, just one trivial change will get the ci-checks build step to pass:

```
--- installer/alpine/build.sh.orig
+++ installer/alpine/build.sh
@@ -7,7 +7,7 @@
 
 build_initramfs() {
 	# Asrockrack kernel module
-	cat > /etc/mkinitfs/features.d/asrockrack.modules <<-EOF
+	cat >/etc/mkinitfs/features.d/asrockrack.modules <<-EOF
 		extra/asrdev.ko
 	EOF
```

joelrebel:
Oh right, I created a new branch with the file in Git LFS instead,
 PR here https://github.com/tinkerbell/osie/pull/286, will close this one

PR #286, opened 2021-09-23
Include the asrockrack bios kernel module

joelrebel:
## Description

This change updates the alpine image to build and include the `asrdev.ko` kernel module,
which enables the Asrockrack BIOS utility `BIOSControl` to set/get BIOS configuration.

## Why is this needed

To export and import BIOS configuration on the Asrockrack hardware.

## How Has This Been Tested?

By building the Alpine docker image

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,4 @@
 docker/lfs/* filter=lfs diff=lfs merge=lfs -text
 installer/alpine/assets-*/* filter=lfs diff=lfs merge=lfs -text
 installer/alpine/eclypsiumdriver-alpine-2.0.0.tgz filter=lfs diff=lfs merge=lfs -text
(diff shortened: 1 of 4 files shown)


ScottGarman:
@joelrebel looks like you added the `.gitattributes` path correctly, but the binary zip file wasn't checked in.

Also, while this otherwise looks like what is needed to add the kernel module, would it also make sense to add the userspace binary as well (I assume there is one) in another commit? In order to test this, I'd like to have something that will invoke the kernel module code.

joelrebel:
Thanks, updated commit pushed with the .zip added into git LFS

jacobweinstock:
Hey @joelrebel, looks like this needs a conflict resolved for `.gitattributes`.

ScottGarman:
Closing this, as I created an alternate PR and cherry-picked these commits for it, PR 294 (which has been merged to main).

PR #287, opened 2021-09-23
Skip NVMe namespace management on more Ampere servers

ScottGarman:
These ones return a system-version of "100" instead of "0100" just to
keep us on our toes.


tstromberg (dismissed):

Raj-Dharwadkar (approved):
Looks good to me.

PR #290, opened 2021-09-27
apps/osie-installer: Configure docker /w syslog log driver

splaspood:
## Description

Configures the alpine local docker to use the 'syslog' logging driver via boots (as derived from the syslog_host kopt).

## Why is this needed
We would like to collect osie-runner output in our central logging systems alongside the rest of the log output from our various services. (boots, hegel, etc)

Once osie's alpine environment has been upgraded to 3.14 it should include a docker capable of 'dual logging' which will allow 'docker logs' to continue working as it would before this change.

## Checklist:

I have:

- [ ] added unit or e2e tests


jacobweinstock:
This seems reasonable. Does anyone else have any concerns or feedback? @mmlb @ScottGarman

stephen-fox:
With the deprecation of OSIE, we will (likely) make this change to Metal's internal fork. It just needs to be re-proposed there.

PR #291, opened 2021-09-30
ensure linux sees all physical drives connected to marvell controller upon reset

mikemrm:
## Description

After clearing the Marvell raid configuration, some time is needed for
the individual drives to register again in linux.

By not waiting, during a deprovision, we may not discover all the
necessary drive to wipe or a wipe may start before the drives are
available to the os.

### DIFF
--- a/docker/scripts/functions.sh
+++ b/docker/scripts/functions.sh
@@ -597,6 +597,26 @@ function marvell_reset() {
 		echo "Marvell-MVCLI - Deleting VD id:$vd"
 		echo y | mvcli delete -f -o vd -i "$vd"
 	done
+
+	# Ensure all physical drives are seen by linux
+	# by matching the serials for the physical drives provided by mvcli
+	# with the block device wwid file
+	pd_serials="$(mvcli info -o pd | awk '/Serial:/ { print $NF }')"
+
+	udevadm settle
+
+	pd_missing=true
+	while $pd_missing; do
+		sleep 1
+		pd_missing=false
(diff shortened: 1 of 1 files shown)


ScottGarman:
Have you been able to test this as a custom osie and see it loop and recover?

mikemrm:
During testing I've seen the code which causes it to loop work correctly, however the way it was looking for serials was not working as expected which is why it loop. Since getting the serials the way it's done now, it has not needed to loop as the mvcli command has been slow enough to return that by the time it's completed the disks are available to the os.

Here is output from the SOL

```shell
++ mvcli info -o pd
++ awk '/Serial:/ { print $NF }'
+ pd_serials='203329F89422
203329F893F9'
+ udevadm settle
+ pd_missing=true
+ true
+ sleep 1
+ pd_missing=false
+ echo 'Marvell-Reset: Checking for drives...'
Marvell-Reset: Checking for drives...
+ for serial in '$pd_serials'
+ grep -q 203329F89422 /sys/class/block/sda/device/wwid /sys/class/block/sdb/device/wwid
+ for serial in '$pd_serials'
+ grep -q 203329F893F9 /sys/class/block/sda/device/wwid /sys/class/block/sdb/device/wwid
+ false
```

mmlb:
I wonder if this would apply to other raid cards too?

mikemrm:
> I wonder if this would apply to other raid cards too?

Wouldn't surprise me

ScottGarman (approved):

PR #292, opened 2021-09-30
Log error when disk is not detected

Raj-Dharwadkar:
Issue: deprovisioning fails with this message "disks[@]: unbound variable" if no
disks are detected.

## Why is this needed

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/docker/scripts/deprovision.sh
+++ b/docker/scripts/deprovision.sh
@@ -71,6 +71,12 @@ function autofail() {
 }
 trap autofail EXIT
 
+set_autofail_stage "checking for disk drives"
+if [[ ${#disks[@]} -eq 0 ]]; then
+	echo "Error: No disk drives detected"
+	exit 1
+fi
+
 # Check BIOS config and update if drift is detected
 if [[ $arch == "x86_64" ]] && [[ $reserved != "true" ]]; then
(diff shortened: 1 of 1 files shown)


docker/scripts/deprovision.sh
```diff
@@ -52,6 +52,11 @@ else
 	echo "Notice: found hardware_id ${hw_id} for instance ${id} in facility ${facility}"
 fi
 
+if [[ ${#disks[@]} -eq 0 ]]; then
```
ScottGarman: This should be moved below where the autofail() trap function is defined and enabled. Then you can set an autofail message that will be shown as the reason for the deprov failure in the API and in our Slack channels. E.g:

set_autofail_stage "ensure at least one disk is found"
(then do your disk check and exit 1 if not found)

That way, when your exit 1 is processed, the API and staff will be shown the message: "Error during ensure at least one disk is found"

ScottGarman (changes requested):

docker/scripts/deprovision.sh
Raj-Dharwadkar: modified to keep it consistent with other message in the script

ScottGarman (approved):

PR #293, opened 2021-10-01
Fix alpine kernel builds and upgrade eclypsium

ScottGarman:
Rebuilding the kernel has been broken with an error when we rebuild a pinned version of the linux-firmware package, due to a 404 error when downloading firmware from an upstream source (cypress). This fixes the alpine kernel builds by un-pinning the linux-firmware package.

This PR then upgrades the eclypsium kernel module and userspace application. We're looking to update to Alpine 3.14 in the near future, and the previous version of the eclypsium kernel module would not build for that kernel.


docker/Dockerfile
```diff
@@ -160,8 +160,8 @@ RUN if [ "$(uname -m)" = 'x86_64' ]; then \
 COPY lfs/mlxup-* /tmp/osie/
 RUN install -m755 -D /tmp/osie/mlxup-$(uname -m) /opt/mellanox/mlxup && rm -rf /tmp/osie/
 
-ARG ECLYPSIUM_AGENT_VERSION=2.0.0
```
mmlb, line 163: delete the old lfs one?

ScottGarman: Deleted.

.gitattributes
```diff
@@ -1,3 +1,4 @@
 docker/lfs/* filter=lfs diff=lfs merge=lfs -text
 installer/alpine/assets-*/* filter=lfs diff=lfs merge=lfs -text
 installer/alpine/eclypsiumdriver-alpine-2.0.0.tgz filter=lfs diff=lfs merge=lfs -text
```
mmlb: here too, maybe change to `installer/alpine/eclypsiumdriver-alpine* ...`?

mmlb (changes requested):

.gitattributes
ScottGarman: Yeah, good call. Fixed.

mmlb (approved):

tstromberg (approved):

PR #297, opened 2021-10-25
osie/alpine: blacklist ipmi modules

dustinmiller:
## Description

Lets blacklist these two IPMI related modules for faster boot times

## How Has This Been Tested?

Drone, manually

### DIFF
--- a/docker/scripts/deprovision.sh
+++ b/docker/scripts/deprovision.sh
@@ -262,6 +262,11 @@ fi
 
 # Run packet-hardware inventory to update API components and firmware details
 set_autofail_stage "running packet-hardware inventory"
+
+if ! modprobe ipmi_si ipmi_devintf; then
+	echo "Unable to modprobe si_ipmi or ipmi_devintf"
+fi
+
 if ! packet-hardware inventory --verbose --tinkerbell "${tinkerbell}/hardware-components"; then
 	echo "Warning: packet-hardware inventory failed for server ${id} (${class})"
 fi
--- a/rules.mk.j2
+++ b/rules.mk.j2
@@ -145,6 +145,8 @@ build/osie-$v/%.tar.gz: build/%.tar.gz
 build/osie-$v/initramfs-{{platform}}: build/osie-$v-rootfs-{{platform}} installer/alpine/init-{{arch}}
 	$(E) "CPIO     $@"
 	install -m755 installer/alpine/init-{{arch}} build/osie-$v-rootfs-{{platform}}/init
+	echo 'blacklist ipmi_si' >> build/osie-$v-rootfs-{{platform}}/etc/modprobe.d/boot-opt-blacklist.conf
+	echo 'blacklist ipmi_devintf' >> build/osie-$v-rootfs-{{platform}}/etc/modprobe.d/boot-opt-blacklist.conf
 	(cd build/osie-$v-rootfs-{{platform}} && find -print0 | bsdcpio --null --quiet -oH newc | pigz -9) >$@.osied
 	install -D -m644 $@.osied $@
 	touch $@


tstromberg (commented):
Conceptually this seems fine. @splaspood @ScottGarman - will this break anything? 

The only thing I could think of is if we did any sort of IPMI-based data collection.

ScottGarman:
We also refer to the `ipmi_defintf` module in the osie installer init script:  https://github.com/tinkerbell/osie/blob/main/installer/alpine/init-x86_64#L390

I don't have good context on why this was added in the first place, so I can't confidently say one way or another whether this will break things. Are we only doing this for faster boot times? And if so, how much is this saving us?

gemagomez:
> I don't have good context on why this was added in the first place, so I can't confidently say one way or another whether this will break things. Are we only doing this for faster boot times? And if so, how much is this saving us?

This is a request that comes from ops that we want to fix this because it is interfering with optimal operation of the cloud. There is no reason to have these servers stuck on probing is they are not going to achieve it due to this hardware not allowing for it. It improves our provisioning and deprovisioning times for these servers by +100 secs.

dustinmiller:
> Conceptually this seems fine. @splaspood @ScottGarman - will this break anything?
> 
> The only thing I could think of is if we did any sort of IPMI-based data collection.

Yes, in deprovision.sh we do use packethost/packet-hardware which collects some information via ipmitool.
In my testing, we're still able to modprobe these and proceed

I split this into two commits for clarity :slightly_smiling_face:

dustinmiller:
> We also refer to the `ipmi_defintf` module in the osie installer init script: https://github.com/tinkerbell/osie/blob/main/installer/alpine/init-x86_64#L390
> 
> I don't have good context on why this was added in the first place, so I can't confidently say one way or another whether this will break things. Are we only doing this for faster boot times? And if so, how much is this saving us?

Before:

```
[    7.029341] microcode: CPU56: patch_level=0x08301034
[    7.034430] microcode: CPU57: patch_level=0x08301034
[    7.039512] microcode: CPU58: patch_level=0x08301034
[    7.044600] microcode: CPU59: patch_level=0x08301034
[    7.049688] microcode: CPU60: patchtate machine
[  118.024500] ipmi_si IPI0001:00: There appears to be no BMC at this location
[  118.034536] ipmi_ssif: IPMI SSIF Interface driver
[  118.045270] Broadcom NetXtreme-C/E driver bnxt_en v1.10.0
````

After:

```
[    9.350745] microcode: CPU41: patch_level=0x08301034                                                                                                                                       
[    9.355827] microcode: CPU42: patch_level=0x08301034                                                                                                                                       
[    9.360912] microcode: CPU43: patch_level=0x08301034                                                                                                                                       
[    9.365992] microcode: CPU4ask: 0xffffffffffffffff max_cycles: 0x240937b9988, max_idle_ns: 440795218083 ns                                                                                 
[    9.592023] clocksource: Switched to clocksource tsc                                                                                                                                       
Alpine Init v3.4.5                                                                                                                                                                            
 * Verifying required kernel cmdline arguments: ignoring packet_state: provisioning                                                                                                           
ok.                                                                                                                                                                                           
 * Loading boot drivers: [    9.762091] loop: module loaded                                                                                                                                   
[    9.766712] squashfs: version 4.0 (2009/01/31) Phillip Lougher                                                                                                                             
ok.                                                                                                                                                                                           
 * Mounting boot media: [    9.799494] Broadcom NetXtreme-C/E driver bnxt_en v1.10.0                                                                                                          
[    9.819694] bnxt_en 0000:01:00.0 eth0: Broadcom BCM57414 NetXtreme-E 10Gb/25Gb Ethernet found at mem f0210000, node addr b0:26:28:f7:d1:32                                                 
[    9.832297] bnxt_en 0000:01:00.0: 63.008 Gb/s available PCIe bandwidth (8 GT/s x8 link)                                                                                                    
[    9.854980] bnxt_en 0000:01:00.1 eth1: Broadcom BCM57414 NetXtreme-E 10Gb/25Gb Ethernet found at mem f0200000, node addr b0:26:28:f7:d1:33                                                 
[    9.867569] bnxt_en 0000:01:00.1: 63.008 Gb/s available PCIe bandwidth (8 GT/s x8 link)                                                                                                    
ok.                                                                                                                                                                                           
 * Obtaining IP via DHCP (eth0)...: 
````

so that seems better

**Comment by dustinmiller** [comment]
Doesn't seem like the test_vm environments will support the modprobe ipmi_si

```

(Btest_deprovision│(B+ modprobe ipmi_si ipmi_devintf
(Btest_deprovision│(Bmodprobe: ERROR: could not insert 'ipmi_si': No such device
```

ScottGarman:
@mmlb do you have any ideas for how we can get the kvm tests working with modprob'ing this kernel module?

tstromberg:
I'd be very very surprised if KVM can  plumb impi devices through :( We may
have to make the logic specific to individual hardware platforms and/or
detect KVM.

mmlb:
> @mmlb do you have any ideas for how we can get the kvm tests working with modprob'ing this kernel module?

Presumably it worked before because the kernel just wouldn't detect the hw and thus not try and modprobe it right?

I don't feel super happy about adding the KVM detection logic. I think this would be the first bit of code in the osie container images that acts differently due to being run in CI, I would love to avoid this if possible.

Instead of modprobing the driver could you try deleting the ipmi blacklist files and calling udev only if deprovisioning instead? It would be nice to let the kernel/udev do their thing.

tstromberg (approved):

PR #298, opened 2021-10-29
Update curl requests to retry 7x (~127 seconds)

tstromberg:
Signed-off-by: Thomas Stromberg <t+github@stromberg.org>

## Description

Update curl calls to retry 7x, or just long enough to survive a 2-minute outage (not including other connection overhead).

This defines and uses an `rcurl` alias where convenient, but in some locations, such as where scripts are written to disk, it uses `--retry 7`. Both behave equivalently.

## Why is this needed

In production, we've occasionally seen instances where metadata servers are busy and/or otherwise unavailable. This will allow provisioning jobs to survive these momentary availability blips and transient networking issues.

## How Has This Been Tested?

It hasn't.

### DIFF
--- a/docker/scripts/osie.sh
+++ b/docker/scripts/osie.sh
@@ -147,7 +147,7 @@ if [[ -z ${cpr_url} ]]; then
 else
 	echo "NOTICE: Custom CPR url found!"
 	echo "Overriding default CPR location with custom cpr_url"
-	if ! curl "$cpr_url" | jq . >$cprconfig; then
+	if ! rcurl "$cpr_url" | jq . >$cprconfig; then
 		phone_home "${tinkerbell}" '{"instance_id":"'"$(jq -r .id "$metadata")"'"}'
 		echo "$0: CPR URL unavailable: $cpr_url" >&2
 		exit 1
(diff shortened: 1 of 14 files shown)


docker/scripts/osie.sh
```diff
@@ -202,10 +202,10 @@ if ! [[ -f /statedir/disks-partioned-image-extracted ]]; then
 		git -C $assetdir checkout FETCH_HEAD
 	elif [[ $image_uri =~ ^https:// ]]; then
 		echo -e "${GREEN}#### Adding custom uri: ${image_uri}${NC}"
-		curl --retry 3 "${image_uri}/image.tar.gz" --output "$assetdir/image.tar.gz"
-		curl --retry 3 "${image_uri}/initrd.tar.gz" --output "$assetdir/initrd.tar.gz"
-		curl --retry 3 "${image_uri}/kernel.tar.gz" --output "$assetdir/kernel.tar.gz"
-		curl --retry 3 "${image_uri}/modules.tar.gz" --output "$assetdir/modules.tar.gz"
+		rcurl "${image_uri}/image.tar.gz" --output "$assetdir/image.tar.gz"
+		ccurl "${image_uri}/initrd.tar.gz" --output "$assetdir/initrd.tar.gz"
```
ScottGarman: Typo?

ScottGarman (changes requested):

docker/scripts/osie.sh
tstromberg: Yes, and fixed. Thanks for noticing it!

ScottGarman (approved):

